Issue #5, opened 2017-12-12
Plans to add Typescript support?


tymondesigns:
Love the simplicity! 

Do you have any plans to add Typescript support at some point?

maybe something like 

```
microbundle --ts
// or
microbundle --transpile=typescript
```

developit:
We could probably build that in by-default, based on file extension.
It looks like [rollup-plugin-typescript2](https://github.com/ezolenko/rollup-plugin-typescript2) would handle this perfectly out-of-the-box.

jbaxleyiii:
I'm happy to take this one. @developit any details you have in mind?

tymondesigns:
Awesome 😁 happy to add myself, if nobody beats me to it

jonstuebe:
Would you be able to add flow support as well?

jaredpalmer:
FWIW I end up using `tsc-watch --onSuccess` to a directory called `compiled`, and then I point rollup there.

developit:
@jonstuebe yup, I'd love a PR to add [rollup-plugin-flow](https://github.com/leebyron/rollup-plugin-flow).

tymondesigns:
Closing because merged :)

PR #6, opened 2017-12-12
feat: add rollup-plugin-flow

maciej-ka:
https://github.com/developit/microbundle/issues/5

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -164,7 +164,7 @@ function createConfig(options, entry, format) {
 			input: entry,
 			external,
 			plugins: [
-				flow({all: true}),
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -163,6 +164,7 @@ function createConfig(options, entry, format) {
 			input: entry,
 			external,
 			plugins: [
+				flow({ all: true }),
```
thejameskyle: I'm not sure that `all: true` should be the default here

dashed: @thejameskyle what are the reasons that `all` shouldn't be default?

PR #10, opened 2017-12-13
Clarify how to specify UMD export name

msfragala:
# What this does
- Adds a `name` option to the CLI, which accepts a string.
- Briefly discusses options for specifying UMB export names in the README

# Context
When bundling UMD build, it is often useful to control the export name. For instance, microbundle defaults to a snake_case version of the `name` field in `package.json`, but camelCase is generally preferred in JavaScript. Trying do so myself, I noticed [the line that determines the name](https://github.com/developit/microbundle/blob/11ee0699fa013d81802322d84c645bf9acd02ef5/src/index.js#L229) feeds from two additional sources:
  - `options.name` which seems like it comes from the CLI options, but no such options exists
  - `pkg.amdName` which is not a standard field AFAIK or mentioned in the README

This PR intends to fix both those things. 🙂

_It's worth noting that adding `--name` from the CLI already feeds to `options.name`. This just makes that fact official, rather than implicit._

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Microbundle includes two commands - `build` (the default) and `watch`. Neither r
 ### `microbundle` / `microbundle build`
 
 By default, microbundle will infer the location of your source entry file
-(the root module in your program) from the `module` field in your `package.json`. It will infer the output directory and filename(s) from the `main` field.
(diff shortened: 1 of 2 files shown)


developit:
FWIW I would like to also switch the default name generation to use camelCase instead of snake_case.

PR #15, opened 2017-12-13
Change default name from snake_case to PascalCase

msfragala:
This changes the default UMD export name from snake_case to PascalCase, per [this comment](https://github.com/developit/microbundle/pull/10#issuecomment-351415227) from another related PR. 🙂

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "acorn-jsx": "4.1.0",
     "asyncro": "^2.0.1",
     "babel-polyfill": "^6.26.0",
+    "camelcase": "^4.1.0",
     "chalk": "^2.3.0",
     "es6-promisify": "^5.0.0",
     "glob": "^7.1.2",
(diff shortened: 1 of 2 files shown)


developit:
Doesn't look like [uppercamelcase](https://unpkg.com/uppercamelcase@3.0.0/index.js) actually does pascalCase - it does `Titlecasing`.  Also I realized - we still need to strip characters that don't work for variable names, since this gets injected into global scope.

I'm thinking something like this:

```js
const safeVariableName = name => name.replace(/(?:^[^a-zA-Z$_]|[^a-zA-Z0-9_$]+)([a-z]?)/g, (s, c) => c.toUpperCase() );
```

Output:

msfragala:
@developit I've updated the `safeVariableName` function to strip out unsafe characters. Regarding `uppercamelcase`, take another look — its output should be PascalCase. It defers the work of camel-casing to [camelcase](https://github.com/sindresorhus/camelcase) and just capitalizes the first character. It worked as expected in my tests.

msfragala:
Good point. I think I had only libraries with PascalCase names in mind, but camelCase is a better default.

developit:
Seems to work well: https://jsfiddle.net/snmfdzb4/

screen shot 2017-12-15 at 7 07 16 pm

PR #16, opened 2017-12-13
adding ts support

tymondesigns:
refs #5 

had a quick stab at it yesterday, although getting an error when running the cli:

image

Any thoughts? @developit


developit:
Funky - looking into this today.

LinusU:
Will this also put a `.d.ts` file in the published package? That seems like something that we should absolutely do...

developit:
@LinusU - think it would make sense to disable the `flow` and `buble` plugins when ts is enabled? We'd just need to tell it to transpile to ES5, but I think that'd garner better output.

LinusU:
Yeah, I think so, TS can compile with ES5 as output, and will then handle ES6/7/etc. syntax

gpoitch:
You could probably drop `buble` completely and have typescript transpile the `.js` files

LinusU:
> Too weird? haha

I like it! 🚀

developit:
@gpoitch I think that would increase output size quite a bit though

KatSick:
@developit is ther any updates on this ?

developit:
Might be worth trying out Babel's typescript support if we switch to Babel.

marvinhagemeister:
Babelsberg TypeScript support is limited in the sense that things like ```const enums``` is not supported. Only the TypeScript compiler has the full feature set for now.

tymondesigns:
It now runs the build and outputs the compiled ts files to dist 🎉  

However getting an annoying output on the console:
image

tymondesigns:
all good now I think @developit

tymondesigns:
🎉

PR #30, opened 2017-12-18
Add postcss and autoprefixer

knipferrc:
Adds postcss rollup plugin and autoprefixer to allow importing css files


knipferrc:
It will match what the main field is in package.json

PR #32, opened 2017-12-20
Migrate CLI to Sade

lukeed:
Some of the "silent" aliases aren't included yet. They aren't documented in the README or in the CLI itself, so I think this may be a good time to address that?

These are the missing aliases:

* `d` --> `output`
* `entries` --> `entry`
* `e` --> `entry`

If these are really intended, `sade@1.1.0` is able to support that need.

I can also move some of the default values from the `index.js` into the CLI directly. For example, passing down `pkg.module` directly, since it's already there... up to you~!

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "bin": "dist/cli.js",
   "scripts": {
     "build": "npm run -s build:babel && npm run -s build:self",
-    "build:babel": "babel-node --presets env src/cli.js --external all --format cjs src/*.js",
+    "build:babel": "babel-node src/cli.js --external all --format cjs src/*.js --presets env",
     "build:self": "node dist/cli.js --external all --format cjs src/*.js",
     "prepare": "npm run -s build",
     "prepare:babel": "babel --presets env src/*.js -d dist && npm t",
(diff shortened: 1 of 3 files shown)


lukeed:
Forgot to include a screenshot of the general help output: 

screen shot 2017-12-20 at 12 37 32 am

developit:
^ liking that help text a lot more!

lukeed:
I think this just needs your input on the "extra" aliases I've omitted for the time being.

Would you like them reincluded?

developit:
I think the entry/entries thing might be necessary unless we're going to bump major.

Does Sade support adding those aliases? Maybe we leave them in for now and then drop them for a 4.0.0?

lukeed:
I can do whatever you want! It didn't initially but it does now.

The only catch is that only 1 alias is displayed per flag...which I don't think matters here since they weren't documented/advertised anywhere lol

developit:
I wonder if it'd be worth manually adding the flags so they don't show up in the docs?

lukeed:
That's what I mean. I'll manually add them via [`parse`](https://github.com/lukeed/sade#progparsearr-opts). They'll work as expected, but the help text won't reflect the additions. It will appear the same it does now (above).

developit:
perfetto 👌

lukeed:
Added in the flag aliases. This is the updated help text:

screen shot 2017-12-24 at 9 04 09 pm

> Only some reordering; no additions

lukeed:
You're probably still on break, but let me know if you need anything here 😄

developit:
its so pretty

Issue #28, opened 2017-12-17
including dependencies


Kikobeats:
Hello,

I'm interested in add support for include all the dependencies in the final bundle.

This conversation started on Twitter:
https://twitter.com/_developit/status/940937797589983235

After check the code, I only seeing a way to excluding, not including:
https://github.com/developit/microbundle/blob/e2bf50b1b5fce002d46b9bf9c1f7c37a74ddd82a/src/index.js#L140

developit:
I believe passing `--inline all` will include all resolvable dependencies from node_modules.

ColinRTaylor:
This PR seems to fix this issue for me: https://github.com/developit/microbundle/pull/33

PR #33, opened 2017-12-20
Include dependencies

ColinRTaylor:
I think this fixes https://github.com/developit/microbundle/issues/28

I ran into this on a project and when it inline anything in "depedencies" will still excluding those in "peerDependencies"

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -139,7 +139,10 @@ function createConfig(options, entry, format) {
 	}
 
 	let useNodeResolve = true;
-	if (options.external==='all' || options.inline==='none') {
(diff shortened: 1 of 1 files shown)


Kikobeats:
Consider review this @developit 😄

Aladdin-ADD:
how to use it? not documented?

I bundled a lib, but seems it didn't include `node_modules/`. --.

ForsakenHarmony:
have you tried `--external none`?

Aladdin-ADD:
@ForsakenHarmony it's not working. and I confirmed the `depedencies` was ok.

PR #38, opened 2018-01-07
Add target environment option

kevlened:
When building isomorphic libraries, it's useful to bundle browser versions of dependencies. The `target` option makes this possible and provides flexibility to make environment-based decisions in the future.

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -13,6 +13,7 @@ prog
 	.option('--entry, -i', 'Entry module(s)')
 	.option('--output, -o', 'Directory to place build files into')
 	.option('--format, -f', 'Only build specified formats', 'es,cjs,umd')
+	.option('--target', 'Specify your target environment', 'node')
 	.option('--external', `Specify external dependencies, or 'all'`)
 	.option('--compress', 'Compress output using UglifyJS', true)
 	.option('--strict', 'Enforce undefined global context and add "use strict"')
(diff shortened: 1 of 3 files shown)


src/cli.js
```diff
@@ -13,6 +13,7 @@ prog
 	.option('--entry, -i', 'Entry module(s)')
 	.option('--output, -o', 'Directory to place build files into')
 	.option('--format, -f', 'Only build specified formats', 'es,cjs,umd')
+	.option('--target', 'Specify your target environment', 'node')
```
kevlened: I use `--target`, because that's [the name in webpack](https://webpack.js.org/configuration/target/). However, `--environment` or `--platform` may be more obvious names.

src/index.js
```diff
@@ -206,7 +206,8 @@ function createConfig(options, entry, format) {
 				}),
 				useNodeResolve && nodeResolve({
 					module: true,
-					jsnext: true
+					jsnext: true,
+					browser: options.target === 'browser'
```
kevlened: Can we just avoid the extra option and set `browser: true` always? If you're building for node, you'll use `--external all` anyway, so this will never be hit.

developit: I'd just do `browser: options.target!=='node'`.

src/cli.js
developit: I'm good with replicating webpack's naming there.

PR #39, opened 2018-01-09
[wip] Babel support

### LINKED ISSUE #25: Doesn't seem to be paying attention to babelrc?
Fooling around: https://github.com/probablyup/markdown-to-jsx/tree/microbundle

Doesn't seem to be honoring my babelrc and it's blowing up on some ES6+ syntax:

```
yarn run v1.3.2
$ microbundle build
SyntaxError: Unexpected token (553:12) /Users/bear/code/markdown/index.js (nodent)
            ...overrideProps,
------------^
✨  Done in 0.75s.
```

Andarist:
Just a starter - need to figure out what else needs to be done to support this within your vision's frame 😉 any particular requirements come to your mind?

fixes #25


developit:
Thanks for the awesome start @Andarist! I think the only things on my mind here were to make sure we were on Babel 7 so we get the updated ultra-compact loose mode and support for blacklisting regenerator (so we can still use nodent or fast-async).

src/index.js
```diff
@@ -168,6 +169,8 @@ function createConfig(options, entry, format) {
 	let cjsMain = replaceName(pkg['cjs:main'] || 'x.js', mainNoExtension);
 	let umdMain = replaceName(pkg['umd:main'] || 'x.umd.js', mainNoExtension);
 
+	const useBabel = !!pkg.babel || await isFile(resolve(cwd, './.babelrc')) || await isFile(resolve(cwd, './.babelrc.js'));
```
developit: I'm thinking that, as long as we can get a default babel config that is as optimized as Bublé, we can just default to Babel.

Andarist: Good stuff happening in terms of babel@7 in terms of outputting better tree-shakeable code & more concise/performant code, but it's still a beta. Im pretty confident in it, we have a comprehensive test suite after all, so I'm using it all over the place - aint sure though if we should promote it just yet as part of other tool (`microbundle` in this context) as it has potential to break some stuff. And also it might not be yet compatible with some other tooling as not every tool had an opportunity to catch up yet and provide support for it.

If we are gonna default to babel@7, Im wondering how do you plan to support babel configs? Or maybe you'd like to ignore it all along and just use internal, optimized one? Im not sure how the whole thing applies to your 0 configuration goal 😉 

Personally I'd say it's better to read the real babel config, but then I'm not sure how we should "optimize" it. Mutating the input and i.e. parsing it to add `loose` option to plugins & presets seems a little bit magical. If one specify his/her config we should just use it as is, right?

Andarist:
> Thanks for the awesome start @Andarist! I think the only things on my mind here were to make sure we were on Babel 7 so we get the updated ultra-compact loose mode and support for blacklisting regenerator (so we can still use nodent or fast-async).

Actually there is a PR for nodent support in babel - https://github.com/babel/babel/pull/7076 . So maybe once it gets merged in, we could just use it as default.

src/index.js
Andarist: Idea - we could possibly support babel config files without any magic, read the config files (only for validation) with some utility exposed by babel like this [one](https://github.com/babel/babel/pull/7201) and parse it (ideally we could just receive the result from babel in JSON) to **warn** people against using non-loose / costly stuff. WDYT?

lrn2prgrm:
Babel has a spetial way to handle named imports that is not spec compliant, I am not sure how bublé works but I just want to be sure you have that on mind in case you plan to always switch to babel (without .babelrc present or any other indication)
Reference:
* https://medium.com/dailyjs/es6-modules-node-js-and-the-michael-jackson-solution-828dc244b8b
* https://medium.com/the-node-js-collection/the-current-state-of-implementation-and-planning-for-esmodules-a4ecb2aac07a

Andarist:
@lrn2prgrm could you explain what do you mean by babel not being spec compliant on named imports? babel is trying to be as spec-compliant as possible by default (when i.e. buble is not trying to do that) - the only pain point is sometimes interop between esm and cjs, but that's because of how both module formats work

abusada:
Hello guys, do you still plan to support babel?

Andarist:
@abusada yes, I very much want this to happen, I'm just slammed with work lately

abusada:
@Andarist maybe i can help, if i know what direction the maintainers would like to go in, i mean, Zero-configuration all the way :D or support `.babelrc`?

i like your idea 

> Idea - we could possibly support babel config files without any magic, read the config files (only for validation) with some utility exposed by babel like this one and parse it (ideally we could just receive the result from babel in JSON) to warn people against using non-loose / costly stuff. WDYT?

and i think it's the best way to move forward with supporting babel!

Andarist:
In general **I think** this is what would be best (some of those poinst I have already stated in this thread, so might be repetitive):
- support babel@7 (drop buble entirely)
- provide default config if user has no `.babelrc`/`.babelrc.js`/`babel.config.js` - there are new APIs in babel@7 for user config inspection
- even if user has custom config then we can "validate" it and suggest i.e. using `loose` mode, or warn about costful transforms
- not yet sure what about nodent

developit:
We'd need to run some tests to see output size of Babel vs Buble for basic stuff.  For async/await there are a few decent babel plugins we could switch to.

developerdizzle:
Any updates on this?  I'd really like to be able to use class properties in my code.

src/index.js
developit: @Andarist I spoke to Henry and I believe he mentioned they pulled the configuration parsing & lookup out of Babel into a standalone module. Maybe we can use that now?

+1 for warning, though it'd be nice to have a way to use a "stock" babel config and let microbundle tweak it into loose mode.

developit:
This is waiting on Babel 7's official release:
https://babeljs.io/blog/2017/12/27/nearing-the-7.0-release

developerdizzle:
@developit Is there any chance of having a pre-release based on 7.0.0-rc.1?

developit:
Maybe!

weslleyaraujo:
that would be really nice, I could test this pre-release in a internal project ☝️

tryzniak:
Will it fix #86 ?

developit:
@tryzniak yes.

developerdizzle:
I'm sure you've heard, but Babel 7 is out https://babeljs.io/blog/2018/08/27/7.0.0

gaui:
So excited for this!

taylorjdawson:
@developit How is the coming along? Really anxious to use babel with microbundle!

ForsakenHarmony:
Just need a default config I'd say

paulmelnikow:
Hi!

I was discussing this project at transitive-bullshit/create-react-library#104, where we're looking for an abstraction over Rollup, and wanted to check in about the status of Babel support in microbundle.

It looks like as of today:

- Bublé is the transpiler
- Babelrc is not respected (#25)
- Babel is being used in a limited way to deal with JSX (#252)
- There is a PR for Babel support, but it's WIP and stale (#39)

I have a couple questions:

1. Have the upstream blocks mentioned above been cleared at this point?
2. If so, would it be helpful if I revived this?

marvinhagemeister:
+1 on reviving this PR. The code in master is already using babel@7, so it should be straightforward to build on that. It doesn't support custom babel configs yet. Note that many babel transforms generate larger output than bublé which is why the latter is used instead of babel's preset-env.

ForsakenHarmony:
Seems like we have a new pr for this

Issue #45, opened 2018-01-12
Wrong flag for strict mode


eLeontev:
For adding strict mode we need to add `--strict=true` flag according the docs.
Expected result:  `use strict` is inserted in the bundle.
Actual result: `use strict` is not inserted in the bundle.

Reasons:  looks like in yargs https://github.com/yargs/yargs/blob/master/docs/api.md#optionkey-opt flag default value of the `type` flag is `'string'` , so your check (`options.strict===true`) always return `false`

Possible solution: add `type: "boolean",` in `yargs.option('strict', { // ...})`
version: 0.2.4

developit:
Should be fixed in the new version, we're not using yargs anymore.

PR #46, opened 2018-01-12
#45 add option's type for strict flag

eLeontev:


developit:
This branch is not up to date with master, we don't use Yargs anymore.

PR #59, opened 2018-01-24
Auto generate tests with snapshots

Andarist:
Just a quick hack to present the idea, this might be fine tuned even more to test automatically for more than just dir structure.


developit:
Ooooh this is perfect! No idea why I didn't go with snapshots in the first place. Thanks!

PR #60, opened 2018-01-24
Override Typescript target if es module

tymondesigns:
This PR overrides the target to be es6 if we are generating an es module


developit:
Hm - we still output ES5 for the ES Modules build, just with ES Modules for the imports/exports.

tymondesigns:
I just think that a consumer may want to keep the es6 intact, e.g. if they only want to support browsers that support es6. and want the smaller files

Not a huge deal though :)

Andarist:
At the moment most applications ignore `node_modules` in their bundlers babel loaders. This way untranspiled code might get to production by accident even if application targets es5 environments.

Personally I wouldnt go this way right now. There are ongoing discussions how to tackle this problem at babel, parcel & CRA boards. Let's wait and see what solution community can agree upon and implement it then.

tymondesigns:
@Andarist yea makes sense 👍

tymondesigns:
closing for now then

PR #65, opened 2018-01-25
Guard against flow transform response

### LINKED ISSUE #64: TypeError: Cannot read property 'code' of undefined
With a module importing `date-fns`, `microbundle` throws "TypeError: Cannot read property 'code' of undefined". See this [reduced test case](https://github.com/tlvince/microbundle-issue-64-reduced-test-case).

tlvince:
Builds started working again fine after this.

Closes #64.

### DIFF
--- a/src/lib/flow-plugin.js
+++ b/src/lib/flow-plugin.js
@@ -5,7 +5,7 @@ export default function fixedFlow(options) {
 	return Object.assign({}, plugin, {
 		transform(code, id) {
 			let ret = plugin.transform(code, id);
-			if (ret.code===code) return null;
+			if (ret && ret.code===code) return null;
 			return ret;
 		}
 	});


Andarist:
Imho it should be investigated first **why** this happens. I've checked it quickly and what gets passed to the flow plugin in your case is:
```js
code === `import { __moduleExports } from "~/Desktop/microbundle-issue-64-reduced-test-case/node_modules/date-fns/is_date/index.js"; export default __moduleExports;`
id === `commonjs-proxy:~/Desktop/microbundle-issue-64-reduced-test-case/node_modules/date-fns/is_date/index.js`
```

It seems like some issue between commonjs and flow plugins.

developit:
seems likely that strip-flow-types is returning `undefined` because there are no flowtype annotations in the file and `{all:true}` is not passed.

PR #68, opened 2018-01-26
Emit bundle events

fwilkerson:
Add the option to emit build events when running microbundle programmatically. This is a feature I'd like to have to notify a connected browser of a change or any build errors that may have occurred.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -99,7 +99,7 @@ export default async function microbundle(options) {
 	}
 
 	if (options.watch) {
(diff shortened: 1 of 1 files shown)


developit:
@fwilkerson awesome! Can I suggest a more explicit API though? Rather than passing `emit`, we can use explicit names (since I don't think there will be many more event types):

```js
microbundle({
  onBuild(e) {
    // handle build completion
  }
})
```

PR #69, opened 2018-01-26
use latest typescript and fix issue with importHelpers

tymondesigns:
This updates the rollup plugin to use the latest version of Typescript, since it [currently uses an ancient version](https://github.com/rollup/rollup-plugin-typescript#typescript-version)

This also fixes an issue where the helpers imported are not transpiled properly
https://github.com/rollup/rollup-plugin-typescript/issues/106

I found this issue when trying to use microbundle to package up a lib that uses the spread operator and fails hard on uglify, since there is a rogue `const` inserted improperly

Obviously it would be better if this were fixed upstream, ([there is a PR for it](https://github.com/rollup/rollup-plugin-typescript/pull/107)) but it's a ghost town over there, hence the many forks of the plugin.


developit:
Makes sense!  I'll merge this and include it in today's release.

Do you think it's worth moving to `rollup-plugin-typescript2`?

tymondesigns:
Great!

I actually already tried that plugin first time round, but it has some issues of its own relating to parallel cache directories conflicting, see [here](https://github.com/ezolenko/rollup-plugin-typescript2/issues/15)
And I tried a few combinations of "fixes" with no luck. Maybe someone else could take a stab at it

PR #70, opened 2018-01-27
Update the `All CLI Options` in the readme

RobinMalfait:
This uses the output of `microbundle --help` from version `0.4.3`

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,31 +50,29 @@ Just like `microbundle build`, but watches your source files and rebuilds on any
 ### All CLI Options
 
 ```
-microbundle [entries..]
+  Usage
+    $ microbundle <command> [options]
 
-Build once and exit
+  Available Commands
+    build    Build once and exit
+    watch    Rebuilds on any change
 
-Commands:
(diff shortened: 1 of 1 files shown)


Andarist:
Could we maybe have this automated with custom script or some package (if there is any package for this)?

RobinMalfait:
It's a nice idea but I don't know of any tools that can do this automatically.
Also, I'm not sure if the output is the same across devices (mac / windows / linux).

Andarist:
> Also, I'm not sure if the output is the same across devices (mac / windows / linux).

I wouldn't worry about this, should be the same across all platforms.

> It's a nice idea but I don't know of any tools that can do this automatically.

Simple custom script should be enough. It would require:
- creating a template for `README.md`
- adding some kind of placeholder line which can be recognized by the script
- creating a script which would - add a header comment with warning that the output file is auto-generated, replacing the mentioned placeholder and outputting the result to the root `README.md`
- adding a precommit hook (i.e with `husky`) which would run this script

Issue #63, opened 2018-01-25
Typescript : should use rollup-plugin-typescript2 instead of rollup-plugin-typescript


freeman:
Great to see the progress on typescript integration !

However the current version uses [rollup-plugin-typescript](https://www.npmjs.com/package/rollup-plugin-typescript) which uses Typescript 1.8 which is very old.

I think switching to [rollup-plugin-typescript2](https://www.npmjs.com/package/rollup-plugin-typescript2) should bring better defaults. In addition I think this should help with #58

developit:
Good catch!

freeman:
Happy to do a PR soon if nobody is on it.

developit:
That'd be awesome @freeman

freeman:
For my use case, seems the rollup-plugin-typescript2 would be a better fit because it supports generating `.d.ts` files from the typescript sources which I consider an essential feature. 

I have tested it and it seems to work perfectly... I'll send a PR and we'll discuss around it ?

Issue #58, opened 2018-01-23
TypeScript declaration files are not emitted


spirosikmd:
I tried the new TypeScript support in version 0.4.0 and it is awesome! Thank you!

However, I have a question regarding the declaration files. Shouldn't they be emitted as well?

*tsconfig.json*
```json
{
  "compilerOptions": {
    "target": "ES2015",
    "module": "commonjs",
    "declaration": true,
    "outDir": "dist"
  },
  "include": ["src/**/*"],
  "exclude": ["node_modules", "**/*-spec.ts"]
}
```

and the output in the `dist` folder contains all files (index.js, index.m.js, index.umd.js and respective map files) except `index.d.ts`. Any idea?

kutyel:
Having the same problem over here! https://github.com/kutyel/linq.ts/pull/107

kutyel:
Plus, @spirosikmd how did you made that work? If I try to target `es2015` I get:
```sh
Error transforming bundle with 'uglify' plugin: Unexpected token: name (List)
```
🤔

developit:
@kutyel using `es2015` will break uglify since it will output classes. Use `"target": "ES3"`.

Regarding the original question - why not place your `index.d.ts` in the repo root?

spirosikmd:
@kutyel es2015 works for me. You can check this [example repository](https://github.com/spirosikmd/typescript-microbundle-example).

@developit when you say to place it in the root, do you mean running the `tsc` command just to output the declaration files? Can't `microbundle` command do this as well?

developit:
ah - I didn't realize this was output from tsc. We should be able to extract it.

kutyel:
@spirosikmd it only works for you because you are not using fully es2015 features (classes, arrows, etc...) if you did you'd get the same error from `uglify` I'm getting. As @developit pointed, the target needs to be `es3`. 👍 

Regarding this issue, yes, it is output that the TypeScript compilers throws and that needs to me added to the *package.json* as `"types": "dist/index.d.ts",` in order to allow people using our modules to use the typings as well 😉

developit:
I believe this should be fixed in [0.4.3](https://github.com/developit/microbundle/releases/tag/0.4.3).

spirosikmd:
I upgraded to 0.4.3 and the declaration files are still not emitted. I am testing with the same `tsconfig.json` as shown in my first comment.

kutyel:
[...]

PR #71, opened 2018-01-28
Switch to `rollup-plugin-typescript2`

freeman:
closes: #63 #58

### DIFF
--- a/test/__snapshots__/index.test.js.snap
+++ b/test/__snapshots__/index.test.js.snap
@@ -41,6 +41,8 @@ Build output to dist:
 exports[`fixtures ts-demo 1`] = `
 "ts-demo
   dist
+    car.d.ts
+    index.d.ts
     ts-demo.js
     ts-demo.js.map
     ts-demo.m.js
@@ -50,6 +52,7 @@ exports[`fixtures ts-demo 1`] = `
   src
     car.ts
     index.ts
+  tsconfig.json
 
 
 Build output to dist:
--- a/test/__snapshots__/index.test.js.snap
+++ b/test/__snapshots__/index.test.js.snap
@@ -52,7 +52,6 @@ exports[`fixtures ts-demo 1`] = `
   src
     car.ts
     index.ts
-  tsconfig.json
 
 
 Build output to dist:
--- a/test/fixtures/ts-demo/src/car.ts
+++ b/test/fixtures/ts-demo/src/car.ts
@@ -1,4 +1,4 @@
-interface Driveable {
+export interface Driveable {
 	drive(distance: number): boolean;
 }
 
--- a/test/fixtures/ts-demo/tsconfig.json
+++ b/test/fixtures/ts-demo/tsconfig.json
@@ -1,0 +1,5 @@
+{
+	"compilerOptions": {
+		"declaration": true
+	}
+}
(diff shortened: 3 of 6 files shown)


test/index.test.js
```diff
@@ -26,10 +26,13 @@ describe('fixtures', () => {
 		}
 
 		it(fixtureDir, async () => {
+			const prevDir = process.cwd();
+			process.chdir(path.resolve(fixturePath));
```
freeman: without this neither `rollup-plugin-typescript` nor `rollup-plugin-typescript2` would find the `tsconfig.json` file in the fixture.

test/fixtures/ts-demo/src/car.ts
```diff
@@ -1,4 +1,4 @@
-interface Driveable {
+export interface Driveable {
```
freeman: without that there was a typescript compilation error

test/index.test.js
```diff
@@ -12,7 +12,7 @@ const constant = konst => () => konst;
 
 const printTree = (nodes, indentLevel = 0) => {
 	const indent = join(times(indentLevel, constant('  ')));
-	return join(nodes.map(node =>
+	return join(nodes.filter(node => node.name[0] !== '.').map(node =>
```
freeman: to prevent `.rpt2-cache` folder from polluting the snapshot.

test/__snapshots__/index.test.js.snap
```diff
@@ -41,6 +41,8 @@ Build output to dist:
 exports[`fixtures ts-demo 1`] = `
 "ts-demo
   dist
+    car.d.ts
+    index.d.ts
```
kutyel: Why does it add **two** declaration files?

freeman: I don't think typescript can generate a single declaration file when there is multiple source files

kutyel: Oh, right, LGTM! 👍

test/fixtures/ts-demo/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+	"compilerOptions": {
+		"declaration": true
```
developit: Any chance we could make this a default?

freeman: It seems difficult to make it the default in a friendly way.

`rollup-plugin-typescript2` only enables me to force it with `tsconfigOverride` but the user would not be able to disable it (which feels wrong...)

I feel like that if you're writing a typescript module, having a `tsconfig.json` is practically a given (because it enables a lot of features in most editors/IDEs) so it is not a big pain.

freeman: made an issue in `rollup-plugin-typescript2` to enable setting the default behaviour in the absence of explicit user intent.

freeman: the feature enabling tsconfig less usage should be coming in the next version of rollup-typescript-plugin2. I feel like we can merge this one (as there is currently no support at all for declaration files) and I'll happily do another PR with the proper defaulting when it becomes available.

Or maybe you prefer to wait a bit @developit ?

freeman:
I upgraded to the latest version of rollup-plugin-typescript2 which enables defaulting the `declaration` bit.

freeman:
@developit any advice on merging this one ? Thanks !

Issue #73, opened 2018-01-28
Disable sourcemap option?


lukeed:
Probably unpopular opinion, but I don't always want sourcemaps to be produced. This is especially true when #66 takes shape.

This isn't a massive pain point or anything, so feel free to disregard~!

Andarist:
I don't find much use for sourcemaps either, so I'd really like to have this too.

PR #79, opened 2018-02-05
Added source-map option

cristianbote:
I was missing this option, and I noticed @lukeed created a issue for it here https://github.com/developit/microbundle/issues/73 and you know, I had a few free minutes and why not, lend a hand. 😄 

Let me know if this is out of scope, or not really something you'd like to go towards.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -310,7 +310,7 @@ function createConfig(options, entry, format, writeMeta) {
 			strict: options.strict===true,
 			legacy: true,
 			freeze: false,
-			sourcemap: true,
+			sourcemap: options.sourcemap===true || options.sourceMap===true,
 			treeshake: {
 				propertyReadSideEffects: false
 			},
--- a/src/index.js
+++ b/src/index.js
@@ -310,7 +310,7 @@ function createConfig(options, entry, format, writeMeta) {
 			strict: options.strict===true,
 			legacy: true,
 			freeze: false,
-			sourcemap: options.sourcemap===true || options.sourcemap===undefined,
+			sourcemap: options.sourcemap!==false,
 			treeshake: {
 				propertyReadSideEffects: false
 			},
--- a/src/index.js
+++ b/src/index.js
@@ -310,7 +310,7 @@ function createConfig(options, entry, format, writeMeta) {
 			strict: options.strict===true,
 			legacy: true,
 			freeze: false,
-			sourcemap: options.sourcemap===true || options.sourceMap===true,
+			sourcemap: options.sourcemap===true || options.sourcemap===undefined,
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -310,7 +310,7 @@ function createConfig(options, entry, format, writeMeta) {
 			strict: options.strict===true,
 			legacy: true,
 			freeze: false,
-			sourcemap: true,
+			sourcemap: options.sourcemap===true || options.sourcemap===undefined,
```
lukeed: This can just be `options.sourcemap` since the default is set to `true` 😄

cristianbote:
Well that was my assumption as well, at first 🙂. Seems that the cli would have it defined from sade as true, but running it through the node package, would end up undefinded.

Did I get it right?

src/index.js
cristianbote: Well that was my assumption as well, at first 🙂. Seems that the cli would have it defined from sade as true, but running it through the node package, would end up undefinded.

Did I get it right?

lukeed: I think it's because your first attempt was using `sourceMap` -- I just checked out your current branch it's working as expected. There's no need for any `===` checks 👍

cristianbote: Ha, seems like @developit did some trickery check there to cover both cases, for default `true` options 😄just checking it that it's not false `options.sourcemap!==false`. That covers both cases, cli and node.

Node it's different, the options object it's not validated by `sade` just a plain config option.

cristianbote: Right about here https://github.com/developit/microbundle/blob/master/src/index.js#L265 I was looking how the other always true values are behaving. And here's how the options are passed to `microbundle` function https://github.com/developit/microbundle/blob/master/test/index.test.js#L29.

lukeed: Those existed for Yargs, aka before the switch to Sade 😄 I didn't bother converting them to make the PR diff easier to follow.

Sade uses [`mri`](https://github.com/lukeed/mri) under the hood, which will _always_ ensure that a flag value will be `boolean|string|number` if specified or if an initial value is provided.

cristianbote: Ha, nice! That makes sense. Also, kudos on the `mri` 😃that's impressive perf, you got there.

lukeed: Haha, thanks 🙃 Flag parsing perf doesn't matter too much for the majority of cases, but is nice to have. The simple utilities of type-casting & assumed booleans are more useful... allowing you avoid all these `!==` manual checks in your apps.

cristianbote:
Hey @lukeed if there's anything else, let me know ;)

developit:
Thanks! Totally needed this today haha

PR #83, opened 2018-02-10
Treat imported subdirectories/files from packages as external too

Andarist:


src/index.js
```diff
@@ -203,7 +204,12 @@ function createConfig(options, entry, format, writeMeta) {
 	let config = {
 		inputOptions: {
 			input: exportType ? resolve(__dirname, '../src/lib/__entry__.js') : entry,
-			external,
+			external: id => {
+				if (options.multipleEntries && id === '.') {
```
Andarist: this is a little bit iffy, but I plan to refactor `multipleEntries` to use `experimentalCodeSplitting` anyway, so ¯\\_(ツ)_/¯

developit: We'll need that under a flag since it won't work in Node.

developit: Just so I know, when is `id==='.'` going to get triggered?

Andarist: > We'll need that under a flag since it won't work in Node.

What do you mean? Was this a comment regarding `experimentalCodeSplitting` or the current setup? As to `experimentalCodeSplitting` - it's not about dynamic imports, just generating N + 1 (I think) files for N entries, where additional file holds all "common" functionality (those used by more than 1 entry)

> Just so I know, when is id==='.' going to get triggered?

I have no idea what's this case unfortunately, I've only moved this condition (hopefully done it right) from above as I had to "merge" new RegExp external test with existing "special case" from [here](https://github.com/developit/microbundle/blob/b20b79bf87f9592209d27094b493a9adb76d70ab/src/index.js#L150)

developit: Ah, ok I had just wondered where it was from.
I'm still catching up on Rollup's code-splitting. It doesn't bundle any loading or transform things to dynamic `import()`?

Andarist: nope, dynamic imports are hidden behind different experimental flag

I've used it [here](https://github.com/rt2zz/redux-persist/pull/705), you can check out what it produces by running build command.

TL;DC (too long, didnt clone 😉 )
For such inputs:
```
experimentalCodeSplitting: true,
input: ['src/index.js', 'src/react.js', 'src/storage.js'],
output: [
  { dir: 'lib', format: 'cjs' },
  { dir: 'es', format: 'es' },
]
```
It produces:
```
es
  - chunk1.js
  - index.js
  - react.js
  - storage.js
lib
  - chunk1.js
  - index.js
  - react.js
  - storage.js
```
where `chunk1.js` has only shared babel helpers.

Rollup is truly amazing at adding 0 extra code when it's not needed. ❤️

Andarist: btw - do you think this PR is ready to merge 😃 ?

lacolaco:
I cannot wait for this feature because I want to use microbundle with `rxjs/*` modules. 😍

Issue #91, opened 2018-02-22
How to consume different bundles from package.json


Kikobeats:
Hey,

I'm intereted in know how to define correctly each bundle at `package.json`.

For example, I know `cjs` bundle is the property `main`.

What happens with the rest? Could be a great resource to put into the readme 🙂

developit:
Definitely worth adding to the readme, yup.

```js
"main": "dist/foo.js",          // CJS bundle
"umd:main": "dist/foo.umd.js",  // UMD bundle
"module": "dist/foo.m.js",      // ES Modules bundle
"source": "src/foo.js",         // custom entry module (same as 1st arg to microbundle)
```

oskarrough:
Thank you for this.

I'm wondering where the `umd:main` property comes from?  
Which bundlers use it? A search didn't get me far… And what about the `browser` field?

linonetwo:
Seems it is a specification from microbuldle and not offical? https://github.com/stereobooster/package.json#microbundle

Kikobeats:
Already on docs 👌

PR #99, opened 2018-03-02
Update README.md - specify output builds in package.json

jvorcak:
It took me a while to figure out how to specify output builds and I think it could be in a README.md

I found out that it is mentioned in #91 so I've made a PR so it's easier to find.


Andarist:
Thanks for the PR! I'm also working on some validation logic that would validate those fields for u, but it's surprisingly unobvious so it will take me few more day to get it done. This is great first step though to make this clearer for people.

Issue #95, opened 2018-02-25
[Question] How to create UMD bundle includings dependencies


Kikobeats:
Hello,

I trying to generate an UMD bundle including all the dependencies necessary. This is the code:
https://github.com/microlinkhq/microlinkjs/blob/master/packages/microlinkjs/src/index.js

I'm running the following command:

```
❯ NODE_ENV=production npx microbundle build --target browser --entry lib/index.js --format es,cjs --inline all
Build output to lib:
      1.54 kB: index.js
      1.54 kB: index.m.js
```

But the filesize is too small, because it doesn't include all the code dependencies.

I'm trying with  `external:none` or `inline:all` but looks like the code doesn't support it
https://github.com/developit/microbundle/blob/master/src/index.js#L152

How can I do that using `microbundle`?

olstenlarck:
I faced that thing too. ANd it was something strage, exactly because CLI flags and options are not one-on-one match.

I believe removing `--target browser` and adding `--external none` would work.

olstenlarck:
I faced that thing too. ANd it was something strage, exactly because CLI flags and options are not one-on-one match.

I believe removing `--target browser` and adding `--external none` would work.

Kikobeats:
@olstenlarck  I tried, but I have the same output:

```
❯ NODE_ENV=production npx microbundle build --entry lib/index.js --format es,cjs --inline all --external none
Build output to lib:
      1.54 kB: index.js
      1.54 kB: index.m.js
```

Andarist:
It seems that this is a bug or a mistake. I think it would be easier to just support `external` with various types of values and we should get rid of `inline` as its both undocumented and not available from the CLI options.

I can work on it later today, but if you need it sooner I encourage you to send a PR for this :)

olstenlarck:
One more thing could be the problem. The plugin `-node-resolve` should be before the `-commonjs`, so that it first resolves the deps, then converts them to ESM if they are CJS.

Kikobeats:
still I didn't find a way to do this ¯\_(ツ)_/¯

jeremenichelli:
In my case, adding `--external none` to the command included packages inside `dependencies` in the package.json file.

PR #107, opened 2018-03-14
Reworked external option slightly (default to all), removed undocumen…

### LINKED ISSUE #106: use --external for a list of specific dependencies
Currently, it seems "all" is the only value to be processed.
When I try to pass parameters to the CLI like `--external classnames,unistore`, `--external classnames, unistore`,  or `--external unistore`,  the output is not changed in either cases.

Andarist:
…ted inline option

this is a **breaking change** ⚠️ 

fixes #106, #95 

I believe defaulting to "all" is way better as we should focus on building libraries with es & cjs formats and bundling deps for that use case is wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Just like `microbundle build`, but watches your source files and rebuilds on any
     -o, --output     Directory to place build files into
     -f, --format     Only build specified formats  (default es,cjs,umd)
     --target         Specify your target environment  (default node)
-    --external       Specify external dependencies, or 'all'
+    --external       Specify external dependencies, or 'none'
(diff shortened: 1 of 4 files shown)


developit:
Reviewed - I think I'm good with this. We'll want to watch for feedback when this goes out, it'll likely change some builds.  My main concern was just that we kept treating `peerDependencies` as the built-in way of denoting that something is external.

Andarist:
I'll look closely to any reported issues with this - I don't expect many though, most likely people won't notice and their builds will get smaller.

The only concern is that it doesn't differentiate umd build by any means.

PR #114, opened 2018-03-28
Support --watch flag

### LINKED ISSUE #112: Support --watch
Hey, first thanks for this awesome project!

It would be nice that the `watch` command is also alternatively exposed as a flag (`--watch`) to ease "composability" of commands:

```json
{
  "build": "microbundle --format cjs --external all --sourcemap false --compress false",
  "start": "yarn build --watch"
}
```

Thanks!

ngryman:
closes #112


Andarist:
Looks good! You just need to update docs [here](https://github.com/developit/microbundle/blob/master/README.md#all-cli-options).

ngryman:
@Andarist Good catch, done!

PR #124, opened 2018-04-22
Source map fix + build only the 2 required files + dep updates

ForsakenHarmony:


Andarist:
@ForsakenHarmony I was not active lately, got slammed with day work. Could I ask why did you close the PR? Is the issue fixed? Is there competing PR for this?

ForsakenHarmony:
oh, whoops

ForsakenHarmony:
meant to close the greenkeeper pr

src/index.js
```diff
@@ -3,7 +3,7 @@ import fs from 'fs';
 import { resolve, relative, dirname, basename, extname } from 'path';
 import chalk from 'chalk';
 import { map, series } from 'asyncro';
-import glob from 'glob';
+import glob from 'tiny-glob/sync';
```
Andarist: any reason why did u chose to use sync version?

ForsakenHarmony: because it was always sync

Andarist:
As this PR reformats stuff quite a lot - I propose to add `prettier` to the project and reformat on precommit hook with `husky` & `lint-staged`. WDYT? Other than that - let's merge it in.

ForsakenHarmony:
yea might be a good idea

ForsakenHarmony:
could add it to this pr as well, but it's already big

can do a new one after jason approves this I suppose

Andarist:
I don't think Jason has much time at the moment to maintain all of those packages that he created 😉I can approve this one (to unlock merging) as well.

The only thing missing at the moment from this PR is a test that would prevent regression. Would it be easy to add one? Or should we just merge it in and worry about it later?

ForsakenHarmony:
Yeah that would also work, what kind of regression test do you mean?

Andarist:
Just some snapshot test (like those that are already here) proving this works now.

ForsakenHarmony:
not sure what specifically you mean, the tests already show that this works

Andarist:
@ForsakenHarmony from what I see no test was added in this PR (I assume those size changes are not related to this PR) - that leads me to the conclusion that there is not test covering this case.

I'm not super attached at the moment to the idea of having test for this as there are dozens cases untested anyway, but would be good to improve this situation - hence my question if it's possible to write some simple (even indirect) test for this. If not - let's just merge it.

ForsakenHarmony:
oh no the size change is not from this pr, so I'd say we can merge this and maybe add those in another pr

PR #130, opened 2018-05-11
Configured prettier

Andarist:
After this gets merged in I propose to follow up it with a commit that would run prettier over configured filed - so new PRs won't contain reformattings. WDYT?


ForsakenHarmony:
Why didn't you just add the lint staged config to `package.json`?

lint-staged.config.js
```diff
@@ -0,0 +1,11 @@
+const jsRule = [
+	'prettier --use-tabs --single-quote --trailing-comma=all --write',
+	'git add'
+];
+
+module.exports = {
+	'src/**/*.js': jsRule,
+	'test/index.test.js': jsRule,
```
Andarist: those rules for `test` dir probably could be rewritten to "every .js file in test but not in test/fixtures" but I always forgot how to do that with glob in combination with `**` :s

Andarist:
> Why didn't you just add the lint staged config to package.json?

Personal preference, can move it to package.json but as I can't use JS there I would be forced to duplicate `jsRule` content among all patterns - IMHO it would be less readable.

Also keeping it out of package.json has a small benefit - we can chose **not to publish** this "config bloat" to npm. It's also worth adding that package.json configs are causing sometimes issues for users - I wouldnt suspect this particular config for lint-staged being source of any trouble, but i.e. babel config in package.json might crash people builds if they use babel@6 (uninstalled plugins/presets might be found there - in `node_modules` and tooling might try to use those incorrectly).

> also it should be precommit, so why didn't this commit already format everything?

`lint-staged` runs specified tasks only on staged files, as this PR didn't touch any the matching files it didn't run tasks on anything

lint-staged.config.js
ForsakenHarmony: I don't think it would be bad to include the fixtures, the only thing we don't want is the outputs

ForsakenHarmony:
> lint-staged runs specified tasks only on staged files, as this PR didn't touch any the matching files it didn't run tasks on anything

right, I forgot, still think we should add that in this pr

lint-staged.config.js
ForsakenHarmony: and I suppose we could use the ignore option for that?

Andarist:
I've tweaked `lint-staged` rules. If you'd still like me to adjust this PR anyhow (not sure if you still want to move its config to `package.json` or not) - let me know.

You can also write down what you'd like me to do & accept this PR with a review. I will adjust this PR accordingly, format all files with those rules (don't want to do it just yet to keep this PR easy to review) & merge in.

ForsakenHarmony:
I don't mind having it separate, it's a balance of polluting `package.json` vs polluting the folder

PR #136, opened 2018-05-21
Turn off eslint rules inconsistent with prettier settings

Andarist:
Alternatively we can use `prettier-eslint` - this would run eslint with `--fix` right after the `prettier`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,15 @@
+{
+  "extends": [
+    "eslint-config-developit",
+    "prettier",
+    "plugin:prettier/recommended"
+  ],
+  "plugins": [
+    "prettier"
+  ],
+  "rules": {
+    "brace-style": "off",
+    "comma-dangle": ["warn", "always-multiline"],
+    "indent": "off"
+  }
+}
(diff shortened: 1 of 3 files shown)


ForsakenHarmony:
I think we could combine the two?

Done that on another project

Andarist:
Hm, do u suggest leaving what I've done here (turning off those eslint rules) and additionally adding `prettier-eslint`? Fine by me, just ensuring I got ur intention right 😃

ForsakenHarmony:
yes and probably also switch lint-staged to `eslint --fix`

developit:
Does this end up enforcing something basically along the lines of semistandard? if so that's fine.

Andarist:
I've adjusted the PR, decided not to use `prettier-eslint` as it's just a thin wrapper around `prettier | eslint --fix`. Using both tools directly seems less awkward.

Andarist:
> Does this end up enforcing something basically along the lines of semistandard? if so that's fine.

Not sure, I'm not familiar with semistandard enough to judge that. This is followup PR to https://github.com/developit/microbundle/pull/130 . I've disabled those rules to suppress eslint errors which got caused by reformatting the codebase with `prettier`.

ForsakenHarmony:
sorry, kinda missed this

PR #137, opened 2018-05-21
Migrate to terser from uglify-js

### LINKED ISSUE #134: Consider migrating to rollup-plugin-terser
`terser` is a new fork of `uglify-es` and the owner's intention is to maintain it, while `uglify-es` (used by our current used version of `rollup-plugin-uglify`) gets deprecated. 

I think it's the right choice to migrate as `terser` handles es6+ syntax.

Andarist:
fixes #134 

~need to first merge #136 and rebase this one on top of that~


developit:
I keep getting bit by Uglify's broken function inlining, which I think is addressed in Terser. Kinda want to try this.

Andarist:
Let's merge it in and see how it plays out. Terser is just a continuation of `uglify-es` (same codebase).

developit:
I keep meaning to run this against some projects and check for size differences before/after.  For me, the `mangle.json` is extremely important (the whole purpose of this bundler tbh), so I want to make sure that keeps working.

Andarist:
Well, there shouldnt be any rly noticeable size changes because this is uglifyjs, just its maintained fork rather than unmaintained one.

> mangle.json

From what i remember thanks to this uglify/terser can reproduce mangled identifiers reliably. Whats the use case for this?

developit:
`mangle.json` isn't for cross-build reliability, it's for static property mappings. You can see what Preact uses this for here:
https://github.com/developit/preact/blob/master/config/properties.json

Essentially, `_properties` are mapped to static names which we treat as private (but still exported and available) API.

matthewmueller:
Oh woops, just saw this. This PR would also fix #201. I accidentally PRed the fix as well in #202

ForsakenHarmony:
is there anything blocking this?

to me it seems like this is a drop in replacement with the same features

developit:
No blockers, I think we can merge.

developit:
I'll merge this weekend. 

@matthewmueller agreed on your note in #202 about actually executing the outputs. We should be able to at least automatically verify the module exports are deep equal to the source's.

developit:
Note: looks like sizes went up? (or at least changed). Might be good to double check what the output looks like with Terser.

ForsakenHarmony:
👀

ForsakenHarmony:
And the tsx fixture wraps the iife in () now

ForsakenHarmony:
I think it's fine

developit:
lol it's literally one semicolon?

matthewmueller:
suhweet!

Issue #141, opened 2018-05-25
Reason support


geoffreydhuyvetters:
Would it make sense to support Reason (.re files) via https://www.npmjs.com/package/rollup-plugin-bucklescript? 
Analog to how we handle TypeScript now.

I think I could create a PR for this.

If not I'm more then happy to do this in a fork, I want to create a small module with Reason as an experiment.

Andarist:
I guess it would make some sense, I have 0 experience with Reason though. Keeping bucklescript compiler as dependency would probably increate the installation size quite a bit, right? Would keeping it as optional dependency work for u? 

If you plan to work on this, please send up a PR and we'll be able to discuss further under it seeing what changes were required to the codebase.

duivvv:
The rollup plugin handles it in the same way, you have to install bsb-platform yourself.

I do think we should provide a default `bsconfig.json` see (https://github.com/shrynx/rollup-plugin-bucklescript#rollup-plugin-bucklescript)

Andarist:
The only thing that I would suggest at this point is to:
- check if there are any reason files in the project directory and use the plugin conditionally
- validate if there is `bsb-platform` installed & throw with a helpful error message if it is not

developit:
That seems like a nice approach.

texastoland:
🤞🏽 I'll take a shot this weekend.

marvinhagemeister:
Closing, looks like there hasn't been much interest in getting this into `microbundle`. We're happy to review PRs if somebody wants to add it :+1:

PR #142, opened 2018-05-25
[FEATURE] [WIP] Add Reason support.

duivvv:
As discussed in https://github.com/developit/microbundle/issues/141.

This PR adds support for [Reason](https://reasonml.github.io/) (`.re`)

It assumes you have `bs-platform` installed (locally or globally) and have a `bsconfig.json` file at the root of your folder. (see fixture)

Just not sure how to tackle the warnings (as suggested in the issue above). 
Any ideas?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ build
 dist
 package-lock.json
 yarn.lock
+test/fixtures/basic-re/lib
+.merlin
--- a/src/index.js
+++ b/src/index.js
@@ -76,7 +76,7 @@ export default async function microbundle(options) {
 	options.name =
 		options.name || options.pkg.amdName || safeVariableName(options.pkg.name);
 
-	const jsOrTs = async filename =>
+	const getFileType = async filename =>
 		resolve(
 			cwd,
 			`${filename}${
@@ -84,7 +84,9 @@ export default async function microbundle(options) {
 					? '.ts'
 					: (await isFile(resolve(cwd, filename + '.tsx')))
 						? '.tsx'
-						: '.js'
+						: (await isFile(resolve(cwd, filename + '.re')))
+							? '.re'
+							: '.js'
 			}`,
 		);
 
@@ -94,8 +96,9 @@ export default async function microbundle(options) {
 			options.entries && options.entries.length
 				? options.entries
 				: options.pkg.source ||
-				  ((await isDir(resolve(cwd, 'src'))) && (await jsOrTs('src/index'))) ||
-				  (await jsOrTs('index')) ||
+				  ((await isDir(resolve(cwd, 'src'))) &&
+						(await getFileType('src/index'))) ||
+				  (await getFileType('index')) ||
 				  options.pkg.module,
 		)
 		.map(file => glob(file))
--- a/src/index.js
+++ b/src/index.js
@@ -18,6 +18,7 @@ import gzipSize from 'gzip-size';
 import prettyBytes from 'pretty-bytes';
 import shebangPlugin from 'rollup-plugin-preserve-shebang';
 import typescript from 'rollup-plugin-typescript2';
+import bucklescript from 'rollup-plugin-bucklescript';
 import flow from './lib/flow-plugin';
 import { readFile, isDir, isFile, stdout, stderr } from './utils';
 import camelCase from 'camelcase';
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -6,3 +6,5 @@ build
 dist
 package-lock.json
 yarn.lock
+test/fixtures/basic-re/lib
```
Andarist: why this one is ignored? does it contain some compiled files?

src/index.js
```diff
@@ -76,15 +77,17 @@ export default async function microbundle(options) {
...
+	const getFileType = async filename =>
 		resolve(
 			cwd,
 			`${filename}${
 				(await isFile(resolve(cwd, filename + '.ts')))
 					? '.ts'
 					: (await isFile(resolve(cwd, filename + '.tsx')))
 						? '.tsx'
-						: '.js'
+						: (await isFile(resolve(cwd, filename + '.re')))
```
Andarist: i would rewrite this `getFileType` to mutliple if+return pairs instead of nesting so many ternaries

src/index.js
```diff
@@ -280,6 +284,7 @@ function createConfig(options, entry, format, writeMeta) {
 	}
 
 	const useTypescript = extname(entry) === '.ts' || extname(entry) === '.tsx';
+	const useReason = extname(entry) === '.re';
```
Andarist: is it unusual to mix `.js` & `.re`? I mean - maybe we should search whole source/src directory with glob to see if there is any `.re` in there?

Andarist: If it's unusual then we should only check the main entry point, like u did

src/index.js
```diff
@@ -331,6 +337,7 @@ function createConfig(options, entry, format, writeMeta) {
 						},
 					}),
 					!useTypescript &&
```
Andarist: seems like we could create a single variable with enumerable values, smth like `language`? that we could reuse in all those conditional places here, seems slightly better than creating a boolean variable for each "language"

Andarist:
As to the proposed warnings - when reason is detected u could do `require.resolve('bs-platform')` and check if it's there and maybe later validate it's version if rollup plugin doesn't support all of them.

.gitignore
duivvv: It’s the downloaded lib when running the compilation for the first time

src/index.js
duivvv: Will do, wasn’t too sure on what to refactor, could use some cleanup indeed

src/index.js
developit: that'd be a nice trick - we wouldn't have to crawl for .re files at all. it's either a Reason project or it isnt.

duivvv:
Gonna further look into it this afternoon.

src/index.js
duivvv: Well, I think Reason / BuckleScript is designed to run side by side with vanilla JS code.

What would the preferred approach be?

- .re as entry point => reason project
- make mixing possible?

Andarist: Kinda depends on how reason is uses (I had yet no chance to play with it, so cannot say from experience) - if mixing is popular we can allow it, if not then do it right now. We can also assume at the moment that we do not support mixing and revisit later if somebody requests it.

Andarist:
> The reason community tends to promote compiling your JavaScript side by side (using .b.js with your .re files and checking them in via VC.

Are they checking in build artifacts? I don't think we can support keeping JS files side by side with `.re` because `microbundle`'s goal is to output flat bundles (although technically `rollup` allows preserving directory structure since some version).

duivvv:
https://bucklescript.github.io/docs/en/interop-overview.html#build-system

I guess we could initially go for 

-> compile -> bundle as if it was ES... or TypeScript project
- .re entry point -> Reason project

Does that sound ok for now?

Andarist:
Fine by me 😉

src/index.js
texastoland: Reason can definitely be mixed. My current use case is to slowly migrate a TypeScript project.

src/index.js
texastoland: I want to attempt this after #198. My feel is the extension checks lack any encapsulation leading to #193 and #196. May I refactor in for #141 or separately? Have you already thought about that @Andarist @developit?

duivvv: @texastoland Feel free to continue on my PR if it makes sense. (not finding a lot of time lately due to multiple factors)

duivvv:
Can't find any time to work on it 😞.
Feel free to re-open if someone wants to continue on it.

ForsakenHarmony:
you don't have to close it 🙃

PR #151, opened 2018-06-11
Updated rollup to 0.60.1, made code changes to support it

Andarist:
closes #146

Still have to figure out what's going with .ts/.tsx tests, because they are failing - but I'v fixed rest of them for now.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "gzip-size": "^4.1.0",
     "pretty-bytes": "^4.0.2",
     "regenerator-runtime": "^0.11.1",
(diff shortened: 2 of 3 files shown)


Andarist:
Ok... tests passing. That is unexpected 😂 as mentioned 2 of them were failing on my machine. Need to investigate this.

ForsakenHarmony:
I've found the tests to be inconsistent as well, the size tests that is

developit:
LGTM.

The size tests are inconsistent with dep upgrades for a reason ;)

Andarist:
> The size tests fail with dep upgrades for a reason ;) it's important to know when things get messed up (because buble got updated, etc)

Sure thing, but this branch was failing on me locally - and I had a fresh install of deps. Hadnt yet time to go back to this to double check.

PR #153, opened 2018-06-14
Fix #111 and log errors during watch mode

### LINKED ISSUE #111: watch mode stops without watching
While `microbundle build` works perfectly fine, I'm not sure what to expect from the `watch` mode.

Currently, it gives me this, and then stops and goes back to the terminal input.
```sh
> microbundle watch
Watching source, compiling to dist:
[object Object]
```

mattdesl:
This PR applies the following:

- Fix #111 
- Ensure that errors during `watch` mode are logged in full, rather than logging `[object Object]` in terminal
- Ensure that fatal errors still trigger build failure, but other error codes should not stop the watch mode
- Pull the logging logic out of `cli.js` into `log-error.js` so that it can be used in `cli.js` and `index.js`

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -1,8 +1,8 @@
 #!/usr/bin/env node
-import chalk from 'chalk';
 import microbundle from '.';
 import prog from './prog';
-import { stdout, stderr } from './utils';
+import { stdout } from './utils';
+import logError from './log-error';
 
 const run = opts => {
 	microbundle(opts)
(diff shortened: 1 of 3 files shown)


Andarist:
I've tested it locally and it seems ok, thanks for your PR!

mattdesl:
For now I also have published `@mattdesl/microbundle` on npm which includes this change.

developit:
Releasing!

Issue #12, opened 2017-12-13
Support JSX ? 


wcastand:
Hi, i wonder if it's possible to use JSX in the source code.
This is a test repo : https://github.com/wcastand/test-microbundle
I have a small repo to test it out and right now i get an error from nodent : 

```bash
▲ test-microbundle at master ✔ yarn build
yarn run v1.3.2
$ microbundle --jsx
SyntaxError: Unexpected token (4:2) /Users/willoo/Documents/Projects/test-microbundle/src/index.js (nodent)
  <div>
--^
✨  Done in 0.50s.
```

In the source code you have an options for jsx in buble `options.jsx` but nodent is before buble.

developit:
I think we can pass this to nodent:

```js
nodent({
  parser: {
    plugins: {
      jsx: require('acorn-jsx')
    }
  }
})
```

wcastand:
Is there a way to add this in my project or i make a PR to include jsx parsing by default in nodent ? or we add an options to include jsx ?

developit:
I think we should just support it out-of-the-box.

wcastand:
There was a release just before this was merged.
Any idea when it will be published ?

Thank for your work and reactivity :)

albinekb:
Hi, I cant seem to get JSX to work either, just installed:
```
$ microbundle
SyntaxError: Unexpected token (37:8) /Users/albinekb/Devel/Projects/react-native-net-info-hoc/src/components/withNetInfo.js (nodent)
        <ComposedComponent {...this.props} connection={{ ...this.state }} />
--------^
✨  Done in 1.41s.
```

wcastand:
It wasn't pushed to npm yet :)
the PR was merged 1 hour after the last release so i think he waits to get more change before pushing a new version :)

alexindigo:
Curious if this is out.

developit:
about to be.

alexindigo:
Actually, I’m seeing some errors. 
I cloned this master, ran build and installed into my project. Everything worked as expected, but then I tried to plug in, created bundle into another prohect and it threw with `h is undefined`.
Looks like it compiled `<h1>` into `h(‘h1’, ...` with `h` not being defined anywhere.

alexindigo:
I can provide more details when I get to the computer.

developit:
That seems correct. The default pragma is `h()`, you can change it to anything you want:

```
microbundle --jsx React.createElement
```

alexindigo:
[...]

PR #163, opened 2018-07-07
add the --jsx document

luxp:
Seems there is no document for `--jsx` flag. And it's needed for build react component.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Just like `microbundle build`, but watches your source files and rebuilds on any
     --cwd            Use an alternative working directory  (default .)
     --sourcemap      Generate source map  (default true)
     -h, --help       Displays this message
-    --jsx            Support building React Component (--jsx React.createElement)
+    --jsx            A custom JSX pragma like React.createElement (default: h)
 ```
 ### Specifying builds in `package.json`
 
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Just like `microbundle build`, but watches your source files and rebuilds on any
     --cwd            Use an alternative working directory  (default .)
     --sourcemap      Generate source map  (default true)
     -h, --help       Displays this message
+    --jsx            Support building React Component (--jsx React.createElement)
 ```
 ### Specifying builds in `package.json`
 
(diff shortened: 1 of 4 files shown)


Andarist:
Does it work? I see it's used in the source code where configs are generated, but at the same time I'm surprised because it's not defined [here](https://github.com/developit/microbundle/blob/acc80d0b9f974b4caf4291077ffa253a8fe50a0c/src/prog.js#L17-L27) and I would suspect it not being parsed/passed down correctly. 🤔

luxp:
It just works.
First time see this in #12 .
After dig into the source code, it seems that it is supported by buble. https://buble.surge.sh/guide/#jsx

https://github.com/developit/microbundle/blob/master/src/cli.js#L20
https://github.com/developit/microbundle/blob/master/src/index.js#L346

And I made a simple repo to test.
https://github.com/luxp/microbundle-react-jsx-demo

developit:
This works because it was already being passed to Bublé here:
https://github.com/developit/microbundle/blob/2104c2956859580fd5fcd8fbbab68ec1394f4aec/src/index.js#L353

However, I think we should take the opportunity to also pass it to TypeScript here:

https://github.com/developit/microbundle/blob/2104c2956859580fd5fcd8fbbab68ec1394f4aec/src/index.js#L334

README.md
```diff
@@ -75,6 +75,7 @@ Just like `microbundle build`, but watches your source files and rebuilds on any
     --cwd            Use an alternative working directory  (default .)
     --sourcemap      Generate source map  (default true)
     -h, --help       Displays this message
+    --jsx            Support building React Component (--jsx React.createElement)
```
developit: I'd prefer a wording of `"A custom JSX pragma like React.createElement (default: h)"`

luxp: Agree to this !

weslleyaraujo:
hey folks, is there anything missing for this to be merged? Quite helpful info, took some time to find out `--jsx` works already 😅

luxp:
Hey guys, is there anything missed for this to merged ?

PR #165, opened 2018-07-10
Use .mjs extension for module main output

developit:
Tis the standard and everything is converging on it, time to adopt!


Andarist:
Beside broken tests - LGTM.

PR #174, opened 2018-07-26
Add raw option to display the raw byte count

andrewiggins:
For libraries that want to keep their size down, showing the raw byte size can be useful to evaluate exactly the size difference between implementations. The `--raw` option shows the raw byte size of all bundles.

Open other names as well.


developit:
YUS

src/index.js
```diff
@@ -135,7 +135,7 @@ export default async function microbundle(options) {
 	}
 
 	function formatSize(size, filename, type) {
-		const pretty = prettyBytes(size);
+		const pretty = options.raw ? `${size} B` : prettyBytes(size);
```
developit: @andrewiggins I'm wondering if we could get around even having to add the `--raw` option here - what if we just always show bytes for <5000b libs?

```js
const pretty = size < 5000 ? `${size} B` : prettyBytes(size)
```

kristoferbaxter:
Alternate proposal – always show the raw bytes in addition to the compressed resource sizes?

Good idea @andrewiggins.

src/index.js
kristoferbaxter: In order to get this to work, you'll likely need to move the option read to `getSizeInfo` now.

Also, the size passed to this function is pre-compressed, so I'd recommend getting the size directly from `code` in the function below.

Andarist:
Code-wise it's good. I'm wondering how useful this is at the moment though - what I mean is that `microbundle` doesn't make it easy now to compare previous & actual bundle sizes and you mention that this the main motivation behind this feature. Maybe we could do better by storing some snapshot file with those sizes so they would be easily comparable when viewing git diffs?

developit:
@Andarist I've done that before by grabbing the file sizes prior to running builds, then calculating the differences post-build.

test/fixtures/raw/src/index.js
```diff
@@ -0,0 +1,50 @@
+export const OneKilobyteOfBobRoss = `
+	We might as well make some Almighty mountains today as well, what the heck.
```
developit: this is my favorite new feature

developit:
I've merged - only change we may want to make here is that it seems like the <5000b automatic raw mode should take effect based on gzipped size rather than original size.

PR #180, opened 2018-07-31
Brotli sizes

kristoferbaxter:
Adds brotli size for each resource.

Code is not in a friendly state, but wanted to share early.


ForsakenHarmony:
code seems decent, but you have to update the tests

PR #186, opened 2018-08-08
Supports --globals config

### LINKED ISSUE #184: UMD dependencies
When you're working on UMD build, if you're using an external module, the name microbundle uses to import this module can be different from the one the module truly uses.

For instance, React exports window.React but microbundle wants to use window.react.

Can we change it?

Ayc0:
Add --globals config in the cli that allows to rename globals dependencies.

Fixes #184

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ Just like `microbundle build`, but watches your source files and rebuilds on any
     -w, --watch      Rebuilds on any change  (default false)
     --target         Specify your target environment  (default node)
     --external       Specify external dependencies, or 'none'
+    --globals        Specify globals dependencies, or 'none'
     --compress       Compress output using UglifyJS  (default true)
     --strict         Enforce undefined global context and add "use strict"
     --name           Specify name exposed in UMD builds
--- a/src/index.js
+++ b/src/index.js
@@ -31,6 +31,14 @@ const safeVariableName = name =>
 			.toLowerCase()
 			.replace(/((^[^a-zA-Z]+)|[^\w.-])|([^a-zA-Z0-9]+$)/g, ''),
 	);
+const parseGlobals = globalStrings => {
+	const globals = {};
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -106,7 +106,6 @@ Here's what's coming up for Microbundle:
 
 [MIT](https://oss.ninja/mit/developit/)
 
-
-[Rollup]: https://github.com/rollup/rollup
-[Bublé]: https://github.com/Rich-Harris/buble
-[Nodent]: https://github.com/MatAtBread/nodent-compiler
```
ForsakenHarmony: why, if anything you should correct `Nodent` to `NoDent`

Ayc0: I didn't change this line, I only added the line for globals but the precommit hook includes lint-staged which modified the rest of the README

Ayc0: Yep, at the beginning, I thought it was due to the prettier extension on my VSCode, and so I've desactivated the format.onSave setting on VSCode

But it behaved the same way when I try to commit after modifying the file with nano...

If you want I can recommit it with --no-verify or without lint-staged

ForsakenHarmony: just seems really weird that it would lower case that and not the link above

Ayc0: 🤷‍♂️

developit: documentation.js does some strange markdown injection stuff. as long as it still works it's probably safe to ignore haha

Ayc0:
I implemented the API this way:
`microbundle --globals npmLibrayName=globalName,other=otherGlobal`

cyrilchapon:
OMFG one has to document this. We've been searching for the way to use lodash UMD for 1 hour.

Ayc0:
@cyrilchapon it's written in the README.md `    --globals        Specify globals dependencies, or 'none'
` and in the `--help` command

cyrilchapon:
Yeah, we just missed the "examples" part.

Issue #66, opened 2018-01-25
--compress true - is it a good default?


Andarist:
IMHO it makes debugging harder, its not easy to look into source code of libraries in node_modules (built by microbundle) nor it is easy to debug microbundle itself because its built by itself with --compress true (default).

I guess the question is what is a primary goal & target of this tool? If producing micro libraries then I think files should be left unminified by default, minifying can be done later by subsequent tools - bundlers etc.

developit:
The goal of microbundle is primarily to package tiny modules that are used in a browser context. In this respect, minification is the most logical default.

freeman:
Don’t know if it is possible but maybe only the UMD build should be minified, given that the CJS and ESM version mostly imply the consumer is using a bundler ?

Andarist:
> Don’t know if it is possible but maybe only the UMD build should be minified, given that the CJS and ESM version mostly imply the consumer is using a bundler ?

It's certainly possible and what would be a good fit here. I propose 3 possible values of compress option:
- undefined (treated as auto/default) - compress only umd
- true - compress all
- false - compress nothing

developit:
Another option that might be worth adding is a way to do `--compress pretty`, which would run Uglify, but use `{ output: { pretty: true } }`. That would apply all the nice inlining and transformations, but produce readable multiline output.

Andarist:
Will work on this, once https://github.com/developit/microbundle/pull/62 gets merged (it has improved support for tests) - and we will be able to discuss further under the PR about used approach.

boneskull:
I don't think it's clear enough this is intended for browser usage due to the default "target" being "node" (this option isn't really explained?)  

I kind of assumed it would gracefully handle builtins, but it doesn't.

VladShcherbin:
Came here with the same issue when bundling for node.

> The goal of microbundle is primarily to package tiny modules that are used in a browser context. In this respect, minification is the most logical default.

Really weird to see this when default target is `node`. Maybe readme should be updated with this one to avoid confusion and better understanding of package purpose.

kzc:
A way to opt out of minification would be useful for debugging.

lukeed:
Instead of an all or nothing `boolean`, can take the same approach as `--format` and pass a group of formats to compress.

```sh
$ microbundle src --format cjs,umd,iife --compress umd,iife
#=> CJS is the only unminified
```

Andarist:
[...]

PR #187, opened 2018-08-12
Better* defaults

lukeed:
Addresses comments in #66 and #73. Went ahead with a PR just so that we can see what it'd look/feel like in action.

**_TLDR:_**

* Only enable compression & sourcemaps when desired
* Seems like most rely upon the `--no-*` flags (myself included) which were undocumented anyway
* Update snapshots' file sizes
* Add two fixtures: `basic-with-compress`, `basic-with-sourcemap`
* Update README docs

---

_<sup><b>*</b></sup> Biased_

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,6 @@
 }
 ```
 
-
 ## 📦 Usage
 
 Microbundle includes two commands - `build` (the default) and `watch`. Neither require any options, but you can tailor things to suit your needs a bit if you like.
@@ -42,6 +41,16 @@ Microbundle includes two commands - `build` (the default) and `watch`. Neither r
 
 By default, microbundle will infer the location of your source entry file
 (the root module in your program) from the `source` field in your `package.json`. It will infer the output directory and filename(s) from the `main` field. For UMD builds, microbundle will use a snake case version of the `name` field in your `package.json` for the export name; you can also specify a name via an `amdName` field or the `name` CLI option.
+
+If you'd like to compress your library for distribution, include the `--compress` flag. You may also produce and include source maps by including the `--sourcemap` flag:
+
+```sh
+$ microbundle --compress --sourcemap
+# or
+$ microbundle build --compress --sourcemap
+```
+
+> **Note:** These flags can also be applied to `watch` mode!
 
 ### `microbundle watch`
 
@@ -66,16 +75,17 @@ Just like `microbundle build`, but watches your source files and rebuilds on any
     -i, --entry      Entry module(s)
     -o, --output     Directory to place build files into
     -f, --format     Only build specified formats  (default es,cjs,umd)
-    -w, --watch      Rebuilds on any change  (default false)
+    -w, --watch      Rebuilds on any change
     --target         Specify your target environment  (default node)
     --external       Specify external dependencies, or 'none'
-    --compress       Compress output using UglifyJS  (default true)
+    --compress       Compress output using UglifyJS
     --strict         Enforce undefined global context and add "use strict"
     --name           Specify name exposed in UMD builds
     --cwd            Use an alternative working directory  (default .)
-    --sourcemap      Generate source map  (default true)
+    --sourcemap      Generate source map
     -h, --help       Displays this message
 ```
+
 ### Specifying builds in `package.json`
 
 You can specify output builds in a `package.json` as follows:
@@ -95,7 +105,6 @@ Here's what's coming up for Microbundle:
 - [x] [TypeScript support](https://github.com/developit/microbundle/issues/5)
 - [x] [Flowtype support](https://github.com/developit/microbundle/issues/5#issuecomment-351075881)
 
-
 ## 🔨 Built with Microbundle
 
 - [Stockroom](https://github.com/developit/stockroom) Offload your store management to a worker easily.
@@ -106,7 +115,6 @@ Here's what's coming up for Microbundle:
 
 [MIT](https://oss.ninja/mit/developit/)
 
-
-[Rollup]: https://github.com/rollup/rollup
-[Bublé]: https://github.com/Rich-Harris/buble
-[Nodent]: https://github.com/MatAtBread/nodent-compiler
+[rollup]: https://github.com/rollup/rollup
+[bublé]: https://github.com/Rich-Harris/buble
+[nodent]: https://github.com/MatAtBread/nodent-compiler
--- a/src/index.js
+++ b/src/index.js
@@ -437,7 +437,7 @@ function createConfig(options, entry, format, writeMeta) {
 			legacy: true,
 			freeze: false,
 			esModule: false,
-			sourcemap: options.sourcemap !== false,
+			sourcemap: !!options.sourcemap,
 			treeshake: {
 				propertyReadSideEffects: false,
 			},
--- a/src/index.js
+++ b/src/index.js
@@ -319,7 +319,7 @@ function createConfig(options, entry, format, writeMeta) {
 					postcss({
 						plugins: [
 							autoprefixer(),
-							options.compress !== false &&
+							options.compress &&
 								cssnano({
 									preset: 'default',
 								}),
@@ -381,7 +381,7 @@ function createConfig(options, entry, format, writeMeta) {
 					// 	[`export default ${rollupName};`]: '',
 					// 	[`var ${rollupName} =`]: 'export default'
 					// }),
-					options.compress !== false && [
+					options.compress && [
 						uglify({
 							sourceMap: true,
 							output: { comments: false },
--- a/src/prog.js
+++ b/src/prog.js
@@ -24,7 +24,7 @@ export default handler => {
 		.option('--strict', 'Enforce undefined global context and add "use strict"')
 		.option('--name', 'Specify name exposed in UMD builds')
 		.option('--cwd', 'Use an alternative working directory', '.')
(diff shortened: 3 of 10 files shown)


texastoland:
For documentation I'd find Sade `example()`s more helpful then update the readme with the new `--help`. I'm willing to PR too.

The rename makes sense but I think it changes the common use case from `--no-compress` to `--source-map` (spelled the way I expected but unimportant) instead of _just works_. Since it's a breaking change why not pursue your proposal in https://github.com/developit/microbundle/issues/66#issuecomment-361095723?

```bash
$ microbundle src
#=> UMD is the only one minified with source maps

$ microbundle src --compress cjs --source-map cjs
#=> CJS is the only one minified with source maps

# and maybe if boolean then keep status quo

$ microbundle src --no-compress --no-source-map
#=> all minified with source maps

$ microbundle src --no-compress --no-source-map
#=> none minified with source maps
```

lukeed:
Yeah, examples would be nice(r) -- just wanted to document before it was forgotten. 

That comment was just a suggestion really. I agree with @Andarist's response in that it's a lot of extra processing & isn't really needed at the end of the day. I think compression & mapping should be all or nothing.

I PR'd what I'd personally like to see. Discussion can change it from here, but this is my proposal 😄

developit:
Can we enable compression and sourcemaps for `--target web` by default? I know there's a bunch of people using Microbundle for Node, but it's still primarily used to bundle packages that get used by browsers.

For me, it's less about full minification (whitespace, etc), and more about running Uglify's optimizations on the bundled code. Perhaps instead of disabling compression, the default could be to enable beautification?

developit:
That ignores the unpkg.com use-case. Also lots of folks don't have their bundlers configured to optimize to the same extent. In cases like Preact, nobody on earth has their bundler configured to optimize to the same extent (property minification, constant inlining, etc).  Also it's slower to process 50kb of uncompressed source than it is to process 5kb of compressed source - comments and whitespace still have to be parsed.

lukeed:
I'd be happy with `target` specific settings 👍 

***Web***

* Build: 
  * compress: true
  * sourcemap: true

* Watch
  * compress: false
  * sourcemap: true


***Node***

* Build:
  * compress: false
  * sourcemap: false

* Watch:
  * compress: false
  * sourcemap: false

Those would be the default behaviors; specifying the `--*`/`--no-*` flags would override default setting.

ForsakenHarmony:
But the preact esm dist is completely unminified

Having uglified source sucks when there's a problem in some lib and you're trying to figure out why

Sure there's unpkg.com but how many people actually use that for prod?

ForsakenHarmony:
Maybe we should output both minified and unminified for the web target

texastoland:
> Having uglified source sucks when there's a problem in some lib and you're trying to figure out why

I see each of your perspectives but as a real world use case this point is why I commented on https://github.com/developit/microbundle/issues/66#issuecomment-412369258. My current personal project is Node but I'm typically front end. I've used public CDNs but never for production.

Would it make sense for Preact to distribute both compressed and non-compressed bundles. In that case maybe you could check the output for `.min.js` and run an extra pass? Just an idea.

maraisr:
@ForsakenHarmony @developit bump 🚀

ForsakenHarmony:
I agree with luke, compression should be off by default, it's pretty much on @developit now

developit:
Circling back to try to unblock this:

There are some transforms we currently use Terser for that genuinely can't be left off by default, especially if we produce both minified and unminified outputs. An example of this is property mangling: if we produce `lib.min.js` and `lib.js`, those two files should use the same properties - otherwise different bundler configurations will result in materially different object shapes being returned.

Here's my pitch:  always run code through Terser, but in the "no compress" cases above, disable variable mangling and enable Terser's output.beautify option.

I've been using this for Preact debugging. It's still a nice improvement over debugging minified output, but it means I don't have to write code that needs to account for properties being "maybe" mangled.

Another option would be to simply move property mangling out of Terser and into Babel. If done in combination with moving Babel to transform _bundles_ rather than source files, this would be relatively inexpensive. It would also allow for the un-minified output to preserve all of the semantics of the source modules, but strip comments and apply property compression to ensure the result minifies adequately in various bundlers.

Here's roughly what that plugin would look like:

```js
import { transformAsync } from '@babel/core';

// a rollup output plugin
export class BabelFinalizer {
  async renderChunk(code, chunk) {
    const result = await transformAsync(code, {
      filename: chunk.filename,
      babelrc: false,
      configFile: false,
      plugins: [
        [manglePropertiesPlugin, minifyOptions]
      ]
    });
  }
}

export default function manglePropertiesPlugin({ types: t }) {
  function getMappedName(name, config) {
    if (!config.properties || !config.properties.regex) return;

    const reserved = config.properties.reserved || [];
    if (reserved.includes(name)) return;

    const reg = new RegExp(config.properties.regex);
    if (!reg.test(name)) return;
    
    let mapping = config.props;
    const keys = Object.keys(mapping);
    if (keys.length === 1 && keys[0] === 'props') {
      mapping = mapping.props;
    }
    
    const key = '$' + name;
    if (Object.prototype.hasOwnProperty.call(mapping, key)) {
      return mapping[key];
    }
  }
  return {
    name: 'babel-plugin-mangle-properties',
    visitor: {
      Identifier(path, state) {
        const config = state.opts;
        const parent = path.parentPath;
        
        if (!(
          (t.isObjectProperty(parent) && path.key === 'key') ||
          (t.isMemberExpression(parent) && path.key === 'property')
        )) return;

        const mapped = getMappedName(path.node.name, config);
        if (mapped != null) {
          path.replaceWith(t.identifier(mapped));
        }
      }
    }
  };
}
```

(see [example transform](https://astexplorer.net/#/gist/bbc5bc6c675c38add75bc5d851993c16/59201185d57ec6b17b2752e7ce5d5b298dd3424b))

Andarist:
> those two files should use the same properties - otherwise different bundler configurations will result in materially different object shapes being returned.

How is this a problem though? Bundlers should not pick both of such files at the same time and from what I understand property mangling can only be done for local shapes - otherwise it becomes an unsafe transform as it changes module's API.

developit:
There is no way to accurately determine locality of property usage, short of building a complete partial evaluator like Closure, and even then it can't be done without additional information not present in JS syntax. Originally I started using this technique in Preact only for internal properties, as a way to intentionally make their names unreliable so that folks wouldn't use them however the usage has expanded since then to be more of a direct size optimization technique.

Your second point is the reason I pushed back on this: if property mangling is enabled for one output, it must be enabled for all outputs. The way it's commonly used, property mangling is equivalent to saying "please replace these symbols in my code with shorter ones" - it's naive, and doesn't even take into account the various runtime forms of property access/introspection like `Object.defineProperties`. It's not a pattern that makes sense for 100% of cases, but it should at least produce consistent results.

Longer-term, I would like to find a way to solve this using syntax, so that Microbundle is not changing the shape of objects but simply inlining already-minified property names.

Here are a few options currently in my head:

### 1. use symbols

A new `mangle.symbols` property could perform this, though the result would not be transparent since the compiled properties would be enumerable and their keys strings. However, the original constant definitions could be compiled and exported (as another entry), so that consumers of a given library could import them and use them to access the internal properties as if they were opaque symbols, even though they are compiled to short strings.

```js
export const CHILDREN = Symbol.for('CHILDREN');
export const DOM = Symbol.for('DOM');
export const ORIGINAL = Symbol.for('ORIGINAL');

export function createVNode(type, props, original) {
  const vnode = {
    type,
    props,
    original,
    [CHILDREN]: null,
    [DOM]: null,
    [ORIGINAL]: original
  };
  if (!original) vnode[ORIGINAL] = vnode;
  return vnode;
}
```

### 2. Simple computed property aliases

Changing the semantics of Symbol.for() is scary. Instead, this would be trivial to implement since Terser handles all requisite inlining cases already. As with the first option, exporting the constants makes it possible to use these as more of a "protected" API surface where such a thing is necessary (an example is `preact-devtools`, which is inherently coupled to preact and the only consumer of many otherwise internal properties).

```js
export const $children = '__k';
export const $dom = '__d';
export const $original = '__v';

export function createVNode(type, props, original) {
  const vnode = {
    type,
    props,
    original,
    [$children]: null,
    [$dom]: null,
    [$original]: original
  };
  if (!original) vnode[$original] = vnode;
  return vnode;
}
```

### 3. Introduce a helper

ECMAScript has private class fields, but not private object fields. We have Symbols for that, but Symbols don't compress well and are an unusual API surface. Introducing a faux-runtime helper could provide a way to avoid messing with the semantics of Symbol, while preserving that usage pattern:

```js
import { property } from 'microbundle';

export const $children = property('children');
export const $dom = property('dom');
export const $original = property('original');

export function createVNode(type, props, original) {
  const vnode = {
    type,
    props,
    original,
    [$children]: null,
    [$dom]: null,
    [$original]: original
  };
  if (!original) vnode[$original] = vnode;
  return vnode;
}
```

---

Ultimately, I would love to have something better than any of these, where full escape analysis is done to determine when properties can be minified. However, even a Closure Compiler style analysis pass would still be missing the ability to define an exposed property as being internal.

texastoland:
Hit this again after not using Microbundle in some time:
https://twitter.com/texastoland/status/1504855004473360385?s=20&t=bbl1PuIFC76YV5kwx6u9DA

I can try to PR `--target node`'s behavior and `--no-compress` in the readme (still poorly documented).

PR #190, opened 2018-08-15
Fix incorrect default output filename (fixes #188)

### LINKED ISSUE #188: Default output filename incorrect (I think)
https://github.com/developit/microbundle/blob/2104c2956859580fd5fcd8fbbab68ec1394f4aec/src/index.js#L106-L109

I had `"main": "lib"` in my `package.json` as a shortcut for `lib/index.js`. Am I wrong that naming the output file after the package effectively contradicts `package.json`?

texastoland:
For example if `"main": "lib"` in `package.json` then the output file should be `lib/index.js` for Node module resolution.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -112,7 +112,7 @@ export default async function microbundle(options) {
 		.forEach(file => options.input.push(...file));
 
 	let main = resolve(cwd, options.output || options.pkg.main || 'dist');
-	if (!main.match(/\.[a-z]+$/) || (await isDir(main))) {
+	if (!main.match(/\.[a-z]+$/)) {
 		main = resolve(main, `${removeScope(options.pkg.name)}.js`);
 	}
 	options.output = main;
--- a/src/index.js
+++ b/src/index.js
@@ -111,9 +111,10 @@ export default async function microbundle(options) {
 		.map(file => glob(file))
 		.forEach(file => options.input.push(...file));
 
-	let main = resolve(cwd, options.output || options.pkg.main || 'dist');
+	const defaultMain = `dist/${removeScope(options.pkg.name)}.js`;
+	let main = resolve(cwd, options.output || options.pkg.main || defaultMain);
 	if (!main.match(/\.[a-z]+$/)) {
-		main = resolve(main, `${removeScope(options.pkg.name)}.js`);
+		main = resolve(main, 'index.js');
 	}
 	options.output = main;
 
--- a/test/__snapshots__/index.test.js.snap
+++ b/test/__snapshots__/index.test.js.snap
@@ -147,6 +147,34 @@ Build \\"basic\\" to dist:
 224 B: basic.umd.js.br"
 `;
 
+exports[`fixtures basic-with-main-dir 1`] = `
+"Used script: microbundle
+
+Directory tree:
+
+basic-with-main-dir
+  dist
+    index.js
+    index.js.map
+    index.mjs
+    index.mjs.map
+    index.umd.js
+    index.umd.js.map
+  package.json
+  src
+    index.js
+    two.js
+
+
+Build \\"basicWithMainDir\\" to dist:
+211 B: index.js.gz
+154 B: index.js.br
+211 B: index.mjs.gz
+155 B: index.mjs.br
+298 B: index.umd.js.gz
+233 B: index.umd.js.br"
+`;
+
 exports[`fixtures custom-source 1`] = `
 "Used script: microbundle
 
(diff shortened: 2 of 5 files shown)


developit:
I'm not sure I like the idea of every module running with `index.js` 🤔
Not sure if this is commonplace, but I've been moving away from relying on the node-specific module resolution as part of moving towards adopting spec-compliant ES Modules.

developit:
FWIW I think the output of this PR can be emulated in current microbundle by specifying main in its entirety, and it'll use `index` as the basename for the rest of the files it creates too:

```
"main": "lib/index.js"
```

I think the long-term intent with microbundle was to infer `"main"` type from file extension (`lib/index.umd.js` produces a UMD bundle, `lib/index.js` produces a CJS bundle, `lib/index.mjs` produces an ES Modules bundle).  Specifying a directory path for `main` makes that harder.

I'm wondering if we could tweak the approach in this PR to solve your `lib/` case while preserving Microbundle's existing behavior?  Not sure. Something like this perhaps:

```diff
-	let main = resolve(cwd, options.output || options.pkg.main || 'dist');
+	let main = resolve(cwd, options.output || options.pkg.main || `dist/${removeScope(options.pkg.name)}.js`);
	if (!main.match(/\.[a-z]+$/) || (await isDir(main))) {
-		main = resolve(main, `${removeScope(options.pkg.name)}.js`);
+		main = resolve(main, 'index.js');
	}
```

texastoland:
I was only confused about consistency with `package.json`. @Andarist cleared it up. The package name does seem like a better default for the browser anyway. Thanks!

texastoland:
@Andarist I was rethinking that sounds wrong. Imagine if no matter what directory you required it checked for a file named after your project. Here's a minimal repo (just `package.json` and a JS) that doesn't work.

https://github.com/texastoland/microbundle-issue-188

Renaming the JS file `index.js` fixes it. Here are the resolution rules followed for `$ node ./microbundle-issue-188`.

1. `'./microbundle-issue-188'` begins with `'./'` or `'/'` or `'../'`
2. `LOAD_AS_DIRECTORY('./microbundle-issue-188')`
    a. `'./microbundle-issue-188/package.json'` is a file
    b. Parse `'./microbundle-issue-188/package.json'`, and look for `"main"` field.
3. `LOAD_AS_FILE('./microbundle-issue-188/lib')`
    a. `'./microbundle-issue-188/lib'` is **not** a file
    b. `'./microbundle-issue-188/lib.js'` is **not** a file
    c. `'./microbundle-issue-188/lib.json'` is **not** a file
    d. `'./microbundle-issue-188/lib.node'` is **not** a file
4. `LOAD_INDEX('./microbundle-issue-188/lib')`
    a. `'./microbundle-issue-188/lib/index.js'` is **not** a file (_this_ is what I'm fixing)
    b. `'./microbundle-issue-188/lib/index.json'` is **not** a file
    c. `'./microbundle-issue-188/lib/index.node'` is **not** a file
5. `THROW "not found"`

> In short defaulting to `$npm_package_name.js` makes npm fail to load `main` if it's a directory. N.b. `index.js` is only used when the entry point is a directory.

Concrete example.

texastoland:
Is there any reason to not make `index.js` the default when the output is a directory instead of a file? Did my example make sense that it contradicts Node resolution rules when using `main`?

texastoland:
> FWIW I think the output of this PR can be emulated in current microbundle by specifying main in its entirety

Of course. But if you do specify `"main": "dist"` in Microbundle for example it can't be resolved by Node. I assume that's not desired behavior.

> I'm wondering if we could tweak the approach in this PR to solve your lib/ case while preserving Microbundle's existing behavior?

I had to reread that a few times to get it. That's seems clever. I'll amend shortly.

Andarist:
Oh, I see what's happening now - indeed it seems to me that it has to be fixed. Personally I don't like this bit though - `|| (await isDir(main))`. which is part of current microbundle algorithm. People really often remove their dist directory before the build script & this condition makes the algorithm unstable. That's also kinda part of my initial confusion - I thought `lib.js` would get created and not `lib/$project_name.js`

texastoland:
@Andarist you're totally right. It's already checking for an extension so it can just be removed right? I tried it and all tests are passing.

Andarist:
Well, unfortunately the current test suite is really far from being comprehensive, removing this isDir check would be breaking (fine by me, not sure how others feel)

texastoland:
@Andarist I just realized the logic is incorrect anyway. It should have been `&& (await isDir(main))` (no extension and a directory). I initially changed it but everything broke. In practice it's unused. Pushing now.

texastoland:
I need to recheck the snapshots. They shouldn't have changed at all. The changed tests were cwd not entry.

texastoland:
@developit your idea worked!

texastoland:
Switched to Rollup.

PR #194, opened 2018-08-19
Bugfix/ts named imports (fixes #193)

### LINKED ISSUE #193: Error compiling index.ts
```
Error: Could not load .../src/index.ts.js 
(imported by .../node_modules/microbundle/src/lib/__entry__.js): 
ENOENT: no such file or directory, open '.../src/index.ts.js'
```

1. TS is detected.
2. `./lib/__entry__.js` requires `index.ts`.
3. Node resolution rules add `.js`.
4. Latest `rollup-plugin-typescript2` gives same error.

I traced it to using named exports in TypeScript.

texastoland:
Fixes really confusing crash using named exports in TypeScript. See issue in title. It's linked to @developit's optimization in 8960114faf8758366cc602d7aa6e2e7c048f0775. I don't like my fix.

- Might it also crash using non-Node features like Flow types?
- Would this be better as a Rollup plugin after transpile that changes the input file?
- The regexes (L296) could be combined
    ```regex
    // before:  /\bexport\s*   (let|const|var|async|function\*?)\s*[a-zA-Z_$*] /
    // after:   /\bexport\s+({|(let|const|var|async|function\*?)\s+[a-zA-Z_$*])/
    // format:  
    const regex = (...args) => new RegExp(String.raw(...args).replace(/\s+/g, ''))

    regex`
      \bexport\s+
      ( {
      | ( let
        | const
        | var
        | async
        | function\*?
        )
        \s+[a-zA-Z_$*]
      )
    `
    ```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -285,8 +285,10 @@ function createConfig(options, entry, format, writeMeta) {
 	let nameCache = {};
 	let mangleOptions = options.pkg.mangle || false;
 
+	const useTypescript = extname(entry) === '.ts' || extname(entry) === '.tsx';
+
 	let exportType;
-	if (format !== 'es') {
+	if (!useTypescript && format !== 'es') {
 		try {
 			let file = fs.readFileSync(entry, 'utf-8');
 			let hasDefault = /\bexport\s*default\s*[a-zA-Z_$]/.test(file);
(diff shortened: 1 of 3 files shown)


texastoland:
> Might it also crash using non-Node features like Flow types?

I'm going to add a test regardless.

> Would this be better as a Rollup plugin after transpile that changes the input file?

I'll attempt this if test fails.

texastoland:
Closing in favor of #200.

PR #202, opened 2018-08-22
change rollup-plugin-uglify@3.x to rollup-plugin-terser. fixes #201.

### LINKED ISSUE #201: invalid .m.js file when compiling typescript files with export default class
Not sure if this is a microbundle issue or maybe a rollup or buble issue, but if you change:

**microbundle/test/fixtures/basic-ts/src/index.ts**

```ts
import Car from './car';

let Ferrari = new Car();

export default Ferrari;
```

To:

```ts
import Car from './car';

export default Car;
```

And run `microbundle`, you end up with a `dist/index.m.js` (or `dist/index.mjs`) file that looks like this:

```js
export default function(){function t(){}return t.prototype.drive=function(t){return!0},t}();
```

Unfortunately this is invalid Javascript, it should be:

```js
export default (function(){function t(){}return t.prototype.drive=function(t){return!0},t}());
```

With parenthesis around the function like this: `(function() { ... })()`.

matthewmueller:
The compressor that `rollup-plugin-uglify` was using (`uglify-es`) produced broken code under certain circumstances where `toplevel: true`. You can see an example here: #201

First I tried upgrading to `rollup-plugin-uglify@4.x`, but it looks like they removed `uglify-es`, so now it only compiles ES5 compliant code. 

Following the instructions here, https://github.com/TrySound/rollup-plugin-uglify#rollup-plugin-uglify-, I tried swapping uglify for `rollup-plugin-terser`.

That swap seemmed to fix this syntax error in #201.

---

**Side note:** we should probably be testing the actual outputs rather than the file size changes. I did check to make sure that the outputs were reasonable, but the tests should probably ensure outputs work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "rollup-plugin-sizes": "^0.4.2",
     "rollup-plugin-strict-alias": "^1.0.0",
     "rollup-plugin-typescript2": "^0.13.0",
-    "rollup-plugin-uglify": "^3.0.0",
+    "rollup-plugin-terser": "^1.0.0",
     "sade": "^1.4.0",
     "tiny-glob": "^0.2.0",
     "tslib": "^1.9.0",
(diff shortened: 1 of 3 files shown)


matthewmueller:
closing in favor of: #137

Issue #169, opened 2018-07-18
Watch mode fails with enoent


mattcosta7:
While calling microbundle watch, i receive an error 

`Error: ENOENT: no such file or directory, rename ..../.rpt2_cache/[hash]/code/cache_ => ..../.rpt2_cache/[hash]/code/cache`

build mode succeeds without issue

package.json
```
{
     "main": "dist/index.js",
     "umd:main": "dist/index.umd.js",
     "module": "dist/index.m.js",
     "source": "src/index.ts",
     "typings": "dist/index.d.ts",
     "files": [
       "dist"
     ],
     "scripts": {
          "start": "./node_modules/.bin/microbundle watch",
      },
}
 ....

```

developit:
Hmm - looks like ezolenko/rollup-plugin-typescript2#34. Probably a permissions error or something.
Try running `rm -rf .rpt2_cache`.

Also you shouldn't need the `./node_modules/.bin/microbundle` - npm takes care of that for you an can just do `"microbundle watch"`.

asmockler:
I'm also getting this error–`rm -rf .rpt2_cache` didn't help. Works fine with `build` but not `watch`. Any other thoughts?

Running this script:

```
$ microbundle watch ./src/index.ts -o ./dist
```

And here is the error:

```
Error: ENOENT: no such file or directory, rename '/Users/andy/src/personal/deku/.rpt2_cache/882e09ab1bd282c901202b8e1a4105edd568729b/code/cache_' -> '/Users/andy/src/personal/deku/.rpt2_cache/882e09ab1bd282c901202b8e1a4105edd568729b/code/cache'

    at Object.fs.renameSync (fs.js:732:3)
    at RollingCache.roll (/Users/andy/src/personal/deku/node_modules/microbundle/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:18992:12)
    at TsCache.done (/Users/andy/src/personal/deku/node_modules/microbundle/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:19723:24)
    at Object.ongenerate (/Users/andy/src/personal/deku/node_modules/microbundle/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:20134:21)
    at /Users/andy/src/personal/deku/node_modules/rollup/dist/rollup.js:22261:58
    at Array.map (<anonymous>)
[...]

PR #204, opened 2018-08-26
fix(typescript): add cache folder for each build format

alexbenic:
Solves #169.

Change event triggers multiple concurrent rebuilds. Error happens when multiple ts instances try to access cache files.


eshiota:
Hey, any idea when the next release is coming up including this commit?

PR #213, opened 2018-09-19
Change "module" field from .m.js to .mjs

Loilo:
`microbundle` generates `.mjs` files by default, this should be reflected in the readme.


ForsakenHarmony:
Thanks

Loilo:
My pleasure.

Issue #181, opened 2018-08-01
Feature Request:  replace/define constants


developit:
#### Usage:

```
microbundle --define NODE_ENV=production
```

#### Result:

input:

```js
if (NODE_ENV==='production') {
  console.log('prod');
}
else {
  console.log('dev');
}
```

output with `--define NODE_ENV=production`:

```js
console.log('prod')
```

All we have to do is use [rollup-plugin-replace](https://github.com/rollup/rollup-plugin-replace) to injec the value. Uglify handles the rest.

ForsakenHarmony:
creator makes a feature request 👀

Andarist:
Not sure if you like the idea but ive written this - https://github.com/Andarist/rollup-plugin-dotenv . Inspired by CRA which uses .env files to keep env vars organised

developit:
@Andarist looks nice - it also supports inlining regular env vars?  Could always have both.

developit:
lol @ForsakenHarmony it was more of a "what do y'all think of this?" :P

Andarist:
> @Andarist looks nice - it also supports inlining regular env vars? Could always have both.

It doesn't - mainly because:
- without special casing (like ur `--define` proposal) it's hard to tell which env vars should be used and I didnt want to include everything that can be found on `process.env`
- when bundling libraries I don't want to have `NODE_ENV` replaced (leaving that to consumers, so a library can differentiate between dev/prod code paths)

It does support overriding with env vars what is specified in .env files though.

developit:
Makes sense. That being said, rollup's syntax is pretty much what I'm in need of - I have 2 builds from the same source, but want to define a variable that changes the output.  Basically something like `#ifdef`:

```js
// a bunch of these
if (PRETTY) {
  // whitespace handling stuff here
}
```

```js
microbundle --define PRETTY=true && microbundle --define PRETTY=false
```

Andarist:
if u have bunch of such variables (btw what they are? if thats not a mystery) it's imho good to gather them in .env files (which can act like a presets of different sets of those vars)

in the future i'd like to add special env vars to microbundle (need to rethink how it should work exactly) - such as REACT, PREACT, SERVER, NATIVE, BROWSER - multiple bundles could be outputted based on used `package.json` keys or smth.

[...]

PR #233, opened 2018-10-25
[wip] Replace / define constants 

peterpme:
See #181 

putting this up so ya'll can get an idea of what i'm thinking


prichodko:
This would be a really useful feature, especially for dead code elimination, which is currently not possible, right?

Are tests only missing thing? I would love to help out.

peterpme:
hey @prichodko,

Yes - tests are missing. I went on vacation and my groove has been off. I'd appreciate the help!

PR #245, opened 2018-11-11
Fix importing json and superagent 

### LINKED ISSUE #211: Importing JSON
Hey!
Firstly, I wanted to say, it's really great package.
It seems I can't import JSON though.

```
> microbundle -o dist --format cjs --external child_process,fs,path

(nodent plugin) SyntaxError: Unexpected token (2:5) R:\code\parcel-that\src\test.json (nodent)
  "a": 2
-----^
```
And it wouldn't be a problem, but few guys in my node_modules do it.
```
> microbundle -o dist --format cjs --external child_process,fs,path

Error: Unexpected token

at R:\code\parcel-that\node_modules\.registry.npmjs.org\lcid\1.0.0\node_modules\lcid\lcid.json:2:8

1: {
2:   "af_ZA": 1078,
            ^
```
Am I doing something wrong? Can I somehow configure microbundle to consume JSON?

### LINKED ISSUE #236: Can't use superagent
Hi 👋  I was recently tweeting to @developit about this ([thread](https://twitter.com/luisrudge/status/1053004106565070848)), but I figured it was best to open an issue so we can move this forward with more heads into this discussion.

I'm trying to use microbundle with a new library, but, unfortunately, it crashes because of superagent:

image

We thought this could be due to JSON support (#211), but I'm not sure at this point.

Any help is appreciated. I'm willing to add a PR for this if we figure out what's going on. Thanks 🎉

luisrudge:
fix #211 
fix #236 

Hi friends 👋 I know it sucks to bundle two changes in one PR, but the superagent fix also includes fixing the json import, so I thought I'd give it a shot anyway 😬 
If you want me to split the json fix and the superagent fix, just let me know.

### In this PR
- Added `rollup-plugin-json` to support importing json files
- Added a test to test json importing
- removed the `useNodeResolve` checks, because it was always true
- **Reordered the plugins in order to add the custom node/browser resolvers to run before the typescript compilation** (this is the superagent fix).

### Known issues
After I reordered the plugins, the `build:babel` script started complaining about the `#!/usr/bin/env node` line in the cli.js file. I don't know why that's the case :( I think I'll need some help to figure this one out. For now, I just removed it so it can build and test successfully.


src/cli.js
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env node
```
ForsakenHarmony: why did you remove this?

luisrudge: > After I reordered the plugins, the build:babel script started complaining about the #!/usr/bin/env node line in the cli.js file. I don't know why that's the case :( I think I'll need some help to figure this one out. For now, I just removed it so it can build and test successfully.

ianwalter: @luisrudge Putting this back in and running `yarn build:babel` didn't complain for me.

luisrudge: @ianwalter in my branch?

luisrudge: I'm happy to rollback that if that's the case. Maybe it was an env issue with my machine 🤔

ianwalter: @luisrudge Yes, I pulled down your fork.

luisrudge: Perfect. I rolled back 😬

ForsakenHarmony:
Can you rebase?

ForsakenHarmony:
Will merge in a bit

developit:
I'm thinking there was some sort of a conditional in there that got cut at some point without me noticing haha

Issue #172, opened 2018-07-21
Handling for..of with Sets and Maps


pshihn:
`for..of` in a Set gets transformed to use the `length` property which is not defined.

e.g:
```javascript
const set = new Set(['a', 'b', 'c']);
for (const s of set) {
  console.log(s);
}
```

Gets transformed into:
```javascript
for (var i = 0, list = set; i < list.length; i += 1) {
  console.log(list[i]);
}
```

(This is likely a babel bug)

ForsakenHarmony:
https://buble.surge.sh/guide/#-dangerousforof-

it's not supported by buble, but we might switch to babel at some point

developit:
FWIW I recall the transpiled code for this is hideous

ForsakenHarmony:
probably less with loose transforms

PR #252, opened 2018-11-24
Fix jsx error

### LINKED ISSUE #74: Error: Plugin 'jsx' not found
Hi there, I was super excited to find this project. Unfortunately, I can't get it working. I end up getting this error whenever I do anything:

```
$ npm run prepare
> react-multicast@1.0.0 prepare /Users/aaronjensen/Source/react-multicast
> microbundle

Error: Plugin 'jsx' not found
```

Here's my repro repo: https://github.com/substantial/react-multicast

Maybe I'm just doing something wrong?

marvinhagemeister:
This PR changes quite a few compiler related thing to fix the nasty jsx error. The error is caused by breaking changes in the `acorn` plugin architecture. Being really complex and not maintained anymore I swapped it out in favour of a babel plugin.

- Replace `nodent` with `babel-plugin-transform-async-to-promises`
- Because the above needs babel, I replaced bublé with it (we may be able to keep both bublé and babel, so that babel only does the `async/await` transform)
- Update rollup dependencies

Note that the compiled file sizes got significantly larger. I still need to investigate why that's the case. Perhaps someone else can have a look at it? /cc @developit 

Fixes #74


marvinhagemeister:
The size difference is coming from `babel-plugin-transform-async-to-promises`.

Source:
```js
async function foo(...args) {
  return [await two(...args), await two(...args)];
}

async function two(...args) {
  return args.reduce((total, value) => total + value, 0);
}

foo();
```

master (via nodent):

```js
function n() {
  for (var n = [], r = arguments.length; r--; ) n[r] = arguments[r];
  return new Promise(function(r, t) {
    return r(
      n.reduce(function(n, r) {
        return n + r;
      }, 0),
    );
  });
}
module.exports = function() {
  for (var r = [], t = arguments.length; t--; ) r[t] = arguments[t];
  return new Promise(function(t, e) {
    return n.apply(void 0, r).then(function(u) {
      try {
        return n.apply(void 0, r).then(function(n) {
          try {
            return t([u, n]);
          } catch (n) {
            return e(n);
          }
        }, e);
      } catch (n) {
        return e(n);
      }
    }, e);
  });
};
```

Output from this PR:

```js
const _async = (function() {
  try {
    if (isNaN.apply(null, {})) {
      return function(f) {
        return function() {
          try {
            return Promise.resolve(f.apply(this, arguments));
          } catch (e) {
            return Promise.reject(e);
          }
        };
      };
    }
  } catch (e) {}

  return function(f) {
    // Pre-ES5.1 JavaScript runtimes don't accept array-likes in Function.apply
    return function() {
      var args = [];

      for (var i = 0; i < arguments.length; i++) {
        args[i] = arguments[i];
      }

      try {
        return Promise.resolve(f.apply(this, args));
      } catch (e) {
        return Promise.reject(e);
      }
    };
  };
})();

function _await(value, then, direct) {
  if (direct) {
    return then ? then(value) : value;
  }

  value = Promise.resolve(value);
  return then ? value.then(then) : value;
}

const two = _async(function(...args) {
  return args.reduce((total, value) => total + value, 0);
});

const foo = function(...args) {
  return _await(two(...args), function(_two) {
    return _await(two(...args), function(_two2) {
      return [_two, _two2];
    });
  });
};

foo();
```

babel.config.js
```diff
@@ -0,0 +1,28 @@
...
+				],
+				targets: {
+					chrome: '58',
+					ie: '11',
+				},
+			},
+		],
+	],
+	plugins: [
+		'babel-plugin-transform-async-to-promises',
```
ForsakenHarmony: why not fast-async?

test/__snapshots__/index.test.js.snap
```diff
@@ -20,12 +20,12 @@ basic
 
 
 Build \\"basicLib\\" to dist:
-211 B: basic-lib.js.gz
```
ForsakenHarmony: the size increase is pretty drastic, do you know what causes it?

babel.config.js
marvinhagemeister: Because it also relies on `nodent` and relying on that again would lead to the same JSX error message.

One could explore to make sure JSX is transformed before passing it to nodent and making absolutely sure that nothing else loads a newer `acorn` version, but that seems fragile to me.

test/__snapshots__/index.test.js.snap
marvinhagemeister: See the first comment :)

marvinhagemeister: @rpetrich made an update to his babel plugin, which improves size quite a bit. On top of that I switched back to `bublé` so that `babel` is only used to transform `async/await` and nothing else. Curiously this all together leads to even smaller size than what's in master 🎉

babel.config.js
ForsakenHarmony: Well, we don't need the jsx plugin if we do that first in another transform

test/__snapshots__/index.test.js.snap
ForsakenHarmony: The goal was to move over to Babel completely at some point (#34) to get all of it's transforms, i.e. #172

marvinhagemeister: I'm not sure how that would be possible given our size constraint. I tried various ways to get size down after switching to babel and came to the conclusion that it isn't really possible. The generated output by babel is always larger than that of `bublé` and would easily add `+20b` for even our simple fixtures.

ForsakenHarmony: Hmm :/

Guessing loose mode also doesn't help

ForsakenHarmony: cc @developit

src/index.js
```diff
@@ -358,19 +358,18 @@ function createConfig(options, entry, format, writeMeta) {
...
-							forOf: false,
-						},
-						parser: {
-							plugins: {
-								jsx: true,
-							},
-						},
-					}),
+					// Only used for async await
+					!useTypescript &&
```
rpetrich: Should TypeScript use the babel transform or it's own generator-styled one? Should be able to have TypeScript target ESNext or ES2017, and then feed into babel for it to do JSX and async/await transforms

test/__snapshots__/index.test.js.snap
marvinhagemeister: Yup, in loose mode it makes the arguments iteration in the `h()` function larger by rewriting the `arguments` iteration and creating a second copy :/

src/index.js
marvinhagemeister: Ohh that's a great idea 👍💯 I'll try it out

src/index.js
```diff
@@ -358,19 +358,18 @@ function createConfig(options, entry, format, writeMeta) {
...
+						babel({
+							exclude: 'node_modules/**',
+							plugins: [
+								'@babel/plugin-syntax-jsx',
+								[
+									'babel-plugin-transform-async-to-promises',
+									{ inlineHelpers: true },
```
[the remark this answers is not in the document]
marvinhagemeister: You're right, that check needs to be added too. Without it rollup doesn't treat it as a module 👍

test/__snapshots__/index.test.js.snap
developit: I hadn't thought of using both Bublé and babel - that's actually an ideal solution. We could probably even have Babel apply the class properties transform (most requested feature) but still use Bublé's loose mode for classes!

marvinhagemeister: @developit Added the class properties transform plugin 👍

marvinhagemeister:
@ForsakenHarmony can you retry the build? I can't reproduce the linting error that failed it locally

developit:
@marvinhagemeister just need to merge https://github.com/developit/microbundle/commit/0d7798705b57a044133db390d9b1d565ec648410

src/index.js
marvinhagemeister: Added it 🎉 Setting the target to `es2017` is necessary because TS uses some things that are not supported yet in preset-env as far as I'm aware.

marvinhagemeister:
@ForsakenHarmony All the review comments have been addressed and I squashed the commits. I think the PR is ready for another review 👍

wardpeet:
When trying to use closure-compiler on preact which still uses babel 6 to do it's transformations. Babel fails to transpile these files with this PR because @core/babel 7 is a dependency. It is used by the rollup-babel-plugin which means all non babel 7 compatible plugins will fail.
image

A few things to consider:
1) we might want to do a major upgrade for the next release and warn users that we only support babel 7.
2) Putting babel as a peer dependency can also be an option but we'll have to remove **one dependency to bundle your library** from our feature list.
3) set babelrc option to false as a default but make it overridable through cli.

marvinhagemeister:
@wardpeet Point 3) seems like the best choice for me. We didn't support custom babel configs before, so we should be ok to disable it.

wardpeet:
sweet didn't knew that so if we just put `babelrc: false` we are good to go :)
https://github.com/rollup/rollup-plugin-babel#external-dependencies

marvinhagemeister:
@wardpeet Added that line to the babel plugin. We weren't supporting this earlier because we didn't had babel included in the first place :)

wardpeet:
@marvinhagemeister works like a charm now!

src/index.js
```diff
@@ -355,31 +359,35 @@ function createConfig(options, entry, format, writeMeta) {
...
+								compilerOptions: {
+									target: 'es2017',
+								},
+							},
 						}),
 					!useTypescript && flow({ all: true, pretty: true }),
-					nodent({
+					// Only used for async await
+					babel({
+						babelrc: false,
```
wardpeet: maybe add a comment on why we are doing this? I can imagine someone opening a PR to put this on true.

ForsakenHarmony:
Still wondering how fast-async would compare (it would not have the same problem, because jsx is handled elsewhere in babel)

marvinhagemeister:
Did a few comparisons agaist `fast-async/nodent`. For standard async functions `babel-plugin-transform-async-to-promises` generates smaller output (~10-20b) and runs slightly faster. The latter can only be measured in benchmarks and probably has no real world effects.

Where `fast-async` has an advantage is nested `for`-loops that have an `await` statement some levels deep. In that case the output is significantly larger compared to the other babel plugin (about half the size). Interestingly the code runs only about half as fast as the other plugin. My guess is that it uses more function calls looking at a first glance. But I haven't really dug into what's going on.

Used the example benchmark code that [is linked here](https://github.com/MatAtBread/fast-async#useful-links), a few microbundle test cases and some of my own code.

ForsakenHarmony:
Guessing you've tried all the options of fast async, well I can merge if you want

marvinhagemeister:
That'd be awesome and I'd love if a release could be made in the following days 🎉 The jsx error is ever present on the CI for `htm`...

marvinhagemeister:
@ForsakenHarmony woohoo, thanks for merging 🎉👍

PR #263, opened 2018-12-04
Replace bublé with babel

### LINKED ISSUE #25: Doesn't seem to be paying attention to babelrc?

esphen:
Hello 👋

First of all - thank you so much for this project, it makes following best practices when packaging libraries for npm really easy!

I saw the PR at #39 and thought I should help out by merging in the latest code from master and getting it up to date. However, a lot seems to have changed since #39 was created, so it was actually a lot easier to start from scratch. So that's what I did, and in this branch it now works with babel 7.

Reading [the discussion in #39](https://github.com/developit/microbundle/pull/39/files#r160703934), I interpreted @developit's comment as intent to replace bublé with babel, so that's what this PR does.

> as long as we can get a default babel config that is as optimized as Bublé, we can just default to Babel

The generated configuration uses a default babel config consisting of preset-env and preset-react, alongside the existing plugins for async/await support. If a babel config file is detected, it uses a clean slate. I hope this is a sensible starting point; free to suggest changes or improvements.

I kept the custom shebang plugin even though it references bublé because the build fails without it.

The tests are currently failing due to increased output sizes. I wanted to get some feedback before I update the snapshots. Should we look into reducing them before going forward with this?

A lot of inspiration in this PR was taken from @Andarist in #39. Thanks a lot for the pointers on where to look! Hope it's okay with you that I did it this way.

Fixes #25


ForsakenHarmony:
Yes, size is important that's why we currently only use babel for async await

package.json
```diff
@@ -44,8 +44,9 @@
   "dependencies": {
     "@babel/core": "^7.1.6",
     "@babel/plugin-proposal-class-properties": "7.1.0",
```
Andarist: this probably should also use `^`, to align with other specified deps

src/index.js
```diff
@@ -377,26 +376,46 @@ function createConfig(options, entry, format, writeMeta) {
...
-								require.resolve('@babel/plugin-proposal-class-properties'),
-								{ loose: true },
-							],
-						],
-					}),
+					babel(
+						(await isFile('.babelrc')) || (await isFile('.babelrc.js'))
+							? {
+									// Babelrc exists, allow project to configure babel
+									exclude: 'node_modules/**',
```
Andarist: not sure if that exclude is needed, babel@7 wont try to transpile things in node_modules anyway (without explicitly stating it do so)

src/index.js
```diff
@@ -377,26 +376,46 @@ function createConfig(options, entry, format, writeMeta) {
...
+							  }
+							: {
+									// No babelrc exists, use default config
+									exclude: 'node_modules/**',
+									presets: [
```
Andarist: when using builtin config we should pass `babelrc: false` option to rollup-plugin-babel

Andarist:
Could u paste in formatted output of a single failed snapshot pre & post this PR? We could see where lies the difference in transpiling and decide how to proceed further with this. 

Thanks for your work!

src/index.js
```diff
@@ -377,26 +376,46 @@ function createConfig(options, entry, format, writeMeta) {
...
-								{ inlineHelpers: true, externalHelpers: true },
-							],
-							[
-								require.resolve('@babel/plugin-proposal-class-properties'),
-								{ loose: true },
-							],
-						],
-					}),
+					babel(
+						(await isFile('.babelrc')) || (await isFile('.babelrc.js'))
```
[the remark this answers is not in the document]
Andarist: after a thought - we should make our internal babel config a public `babel-preset-microbundle`, but should be done in a followup review - because it will require setting up a monorepo.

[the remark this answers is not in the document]
marvinhagemeister: ohh I like that idea of a dedicated `microbundle` preset 👍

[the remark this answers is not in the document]
loganfsmyth: @Andarist Could we expose something like https://github.com/babel/babel-loader/#customized-loader from `rollup-plugin-babel`? I think it is very bad to encourage frameworks to implement their own resolution for `.babelrc` files and such, since it doesn't necessarily conform to Babel's own implementation details.

src/index.js
```diff
@@ -377,26 +376,46 @@ function createConfig(options, entry, format, writeMeta) {
...
+									plugins: [
+										[
+											require.resolve(
+												'babel-plugin-transform-async-to-promises',
+											),
+											{ inlineHelpers: true, externalHelpers: true },
+										],
+										[
+											require.resolve(
+												'@babel/plugin-proposal-class-properties',
```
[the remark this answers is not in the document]
esphen: As far as I can tell from [the sources](https://github.com/babel/babel/blob/master/packages/babel-preset-env/data/plugins.json#L237), yes. class-properties, however, are not.

src/index.js
```diff
@@ -226,6 +226,45 @@ export default async function microbundle(options) {
 	);
 }
 
+function createBabelConfig(options) {
+	const partialConfig = loadPartialConfig();
+	const hasBabelConfigFile = partialConfig.hasFilesystemConfig();
```
esphen: Note that this function does not respect the `babel` field in package.json

loganfsmyth: The thing is that `loadPartialConfig` is meant to be called per-file, because some files may have a `.babelrc` and some may not. That is what drove my comment above. It seems like `rollup-plugin-babel` should be exposing this as a feature, which is what we do for `babel-loader` so that people can do the same thing.

esphen: @loganfsmyth: I see, thanks for the clarification. So if we go for that solution, that would mean this is impeded on `rollup-plugin-babel` exposing this feature.

I can make an issue over at rollup-plugin-babel to the best of my ability and see where that takes us.

src/index.js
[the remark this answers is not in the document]
Andarist: > @Andarist Could we expose something like https://github.com/babel/babel-loader/#customized-loader from rollup-plugin-babel?

Sure, gonna look through this to see what it actually is & how it can be used to achieve goals here. 

> I think it is very bad to encourage frameworks to implement their own resolution for .babelrc files and such, since it doesn't necessarily conform to Babel's own implementation details.

Totally agreed - I was only pointing to babel-jest & others to see how similar goal is achieved. It's 100% better to reuse the logic in babel itself to avoid buggy implementations.

esphen:
Here's the list of changes to the compiled output before and after this change. It's formatted with prettier to make it easier to read, so the linked gists will noe match 1-1 with the bytes listed in the table below.

| Filename               | Before | After | Change                            | Before                                                                                                       | After                                                                                                |
| ---------------------- | ------ | ----- | --------------------------------- | ------------------------------------------------------------------------------------------------------------ | ---------------------------------------------------------------------------------------------------- |
| basic-css.js           | 133    | 133   | <span color="black">0%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-css-before-js)           | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-css-js)           |
| basic-json.js          | 168    | 113   | <span color="green">-32.7%</span> | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-json-before-js)          | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-json-js)          |
| basic-lib.js           | 468    | 502   | <span color="red">7.3%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-lib-before-js)           | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-lib-js)           |
| basic-lib-ts.js        | 130    | 95    | <span color="green">-26.9%</span> | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-lib-ts-before-js)        | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-lib-ts-js)        |
| basic-lib-tsx.js       | 304    | 206   | <span color="green">-32.2%</span> | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-lib-tsx-before-js)       | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-basic-lib-tsx-js)       |
| class-decorators-ts.js | 591    | 559   | <span color="green">-5.4%</span>  | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-class-decorators-ts-before-js) | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-class-decorators-ts-js) |
| class-properties.js    | 127    | 127   | <span color="black">0%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-class-properties-before-js)    | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-class-properties-js)    |
| custom-source.js       | 472    | 506   | <span color="red">7.2%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-custom-source-before-js)       | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-custom-source-js)       |
| esnext-ts.js           | 891    | 897   | <span color="red">0.7%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-esnext-ts-before-js)           | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-esnext-ts-js)           |
| jsx.js                 | 294    | 252   | <span color="green">-14.3%</span> | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-jsx-before-js)                 | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-jsx-js)                 |
| name-custom-amd.js     | 474    | 508   | <span color="red">7.2%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-name-custom-amd-before-js)     | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-name-custom-amd-js)     |
| name-custom.js         | 470    | 504   | <span color="red">7.2%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-name-custom-before-js)         | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-name-custom-js)         |
| no-pkg.js              | 465    | 499   | <span color="red">7.3%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-no-pkg-before-js)              | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-no-pkg-js)              |
| no-pkg-name.js         | 470    | 504   | <span color="red">7.2%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-no-pkg-name-before-js)         | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-no-pkg-name-js)         |
| pretty.js              | 12522  | 12522 | <span color="black">0%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-pretty-before-js)              | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-pretty-js)              |
| raw.js                 | 2960   | 2960  | <span color="black">0%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-raw-before-js)                 | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-raw-js)                 |
| shebang.js             | 101    | 101   | <span color="black">0%</span>     | [before](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-shebang-before-js)             | [after](https://gist.github.com/esphen/805e0ccd14e8df37b5a504930b682a63#file-shebang-js)             |

Andarist:
| test case                                                                                                                                           | reason of the diff |
| -------------------------------------------------------------------------------------------------------------------------------------------------- | ------ |
| basic-json | babel has removed iterating through arguments because `args` stayed unreferenced, this for sure is OK in loose mode which we want to use, probably even OK in general case (cant find in my hand a thing that it could break even in very contrived situations) |
| basic-lib | when spreading arguments babel is initializing created array with `new Array(length)`, buble creates `[]`. Allocating upfront is probably a little bit better here, so I would classify it as a good tradeoff. Also - babel iterates from left to right, buble iterated from right to left and leverages the fact that the end condition will be met automatically when using i-- and reaching 0, we could propose this to babel as a minor output optimization. |
| basic-lib-ts | it has untranspiled class, so this is a bug somewhere |
| basic-lib-ts | again untranspiled stuff - classes, arguments rest & arrow function, seems like TS transpilation is broken |
| class-decorators-ts | again untranspiled TS stuff |
| custom-source | same as basic-lib |
| esnext-ts | untranspiled `const` declaration which prevents being able to merge it with other variable declarations there |
| jsx | same reason as basic-json |
| name-custom-amd | same as basic-lib |
| name-custom | same as basic-lib  |
| name-pkg | same as basic-lib  |
| name-pkg-name | same as basic-lib |

esphen:
> again untranspiled stuff - classes, arguments rest & arrow function, seems like TS transpilation is broken

Yes, compilation of TypeScript files was missing from the rollup-plugin-babel extension set. Fixed that now and updated the gist and the output comment with updated file sizes.

TS is now up in size, except for decorators, which is now flat.

One annoyance that popped up now is the failing test, see below. I tried debugging this, but this is a little too far into babel internals for me atm. It seems to be related to the transpilation of the async function in the esnext-test. If anyone could look help me into this, that would be awesome.

```
 fixtures › esnext-ts
    ReferenceError: Container is falsy
      at NodePath._replaceWith (node_modules/@babel/traverse/lib/path/replacement.js:188:11)
      at NodePath._remove (node_modules/@babel/traverse/lib/path/removal.js:51:10)
      at NodePath.remove (node_modules/@babel/traverse/lib/path/removal.js:30:8)
      at Object.VariableDeclaration (node_modules/babel-plugin-transform-async-to-promises/async-to-promises.ts:1010:21)
      at NodePath._call (node_modules/@babel/traverse/lib/path/context.js:53:20)
      at NodePath.call (node_modules/@babel/traverse/lib/path/context.js:40:17)
      at NodePath.visit (node_modules/@babel/traverse/lib/path/context.js:88:12)
      at TraversalContext.visitQueue (node_modules/@babel/traverse/lib/context.js:118:16)
      at TraversalContext.visitSingle (node_modules/@babel/traverse/lib/context.js:90:19)
      at TraversalContext.visit (node_modules/@babel/traverse/lib/context.js:146:19)
```

src/index.js
```diff
@@ -226,6 +226,46 @@ export default async function microbundle(options) {
...
+
+	// Allow project to configure babel with local config file
+	if (hasBabelConfigFile) return {};
+
+	// No babelrc exists, use default config
+	// TODO move this into a separate babel-preset-microbundle package
+	return {
+		babelrc: false,
+		extensions: [...DEFAULT_EXTENSIONS, '.ts', '.tsx'],
```
Andarist: atm `rollup-plugin-typescript2` is responsible for transpiling typescript and we should keep it - especially in this PR. We can decide to transpile TS with babel later on, but it has some caveats so it needs extra consideration.

Andarist: Oh, ok - makes sense. I was not aware of the fact that the transpilation has been splitted into 2 steps.

I see that u have updated the before/after table. Super cool 👏 Did you also update the gists?

esphen: Yep, it's updated :) The following files have changed:
- basic-lib-ts.js 
- basic-lib-tsx.js 
- class-decorators-ts.js
- esnext-ts.js (Removed because test fails)

Andarist:
Additional note - babel wraps transpiled class in an IIFE which increases the size a little bit, BUT by adding #__PURE__ comments to those IIFEs it makes them treeshakable (so its superior in this sense to buble). Unfortunately static assignments, such as assignments to prototypes, deopts most treeshaking/DCE algorithms so its better to avoid those in top level scope.

src/index.js
```diff
@@ -419,15 +440,6 @@ function createConfig(options, entry, format, writeMeta) {
 							};
 						},
 					},
-					buble({
-						exclude: 'node_modules/**',
-						jsx: options.jsx || 'h',
-						objectAssign: options.assign || 'Object.assign',
-						transforms: {
-							dangerousForOf: true,
```
Andarist: atm this PR removes `dangerousForOf` behaviour (thus making the output bigger, but more safe), probably as part of this PR we should keep previous semantics, to do the same with babel you have to use this:
```
["@babel/plugin-transform-for-of", { "assumeArray": true }]
```

src/index.js
esphen: Good point! Babel would not work on typescript files if the user also has a babel config file.

This entire logic may change when we add rollup-plugin-babel custom loader, but let's add it for now.

src/index.js
esphen: Sure thing, added. No difference in the diffs of the tests, but that is to be expected, as the only file using for-of (`esnext-ts.js`) is currently failing

developit: wow, I didn't know Babel had an equivalent for this!

developit:
A bunch of them fall into the 7.2% increase that I think is worth figuring out.  I don't think the Array preallocation is actually beneficial in the case of arguments - Buble was right in doing a backwards loop, since that results in the Array being initialized and immediately set to the correct length (largest index first).

Perhaps we can get Babel to change its output format for rest parameters?

developit:
I've opened a PR to Babel to make the outputs align: babel/babel#9152

rpetrich:
@esphen A good portion of that output is `tslib`'s `__asyncValues` helper which is used to transform for await when TypeScript is configured to target `es2017`; switching to target `esnext` will use `babel-plugin-transform-async-to-promises`'s equivalent helper which is more compact. Also, I plan to publish later this evening.

esphen:
@rpetrich Thanks for the tip. I tried setting typescript's target to `esnext`, and that resulted in reducing the bundle size from 2779 to 2490 bytes (-10.4%). Should also mention that the only test that changed in size was `esnext-ts`.

I would like to hear from the other maintainers before we hand over responsibility of compiling ES from the ts plugin to babel, however. Currently typescript only compiles to ES2017, then bublé/babel does the rest. Any good reason it is done this way currently?

Andarist:
Could you rebase on top of the master?

> I would like to hear from the other maintainers before we hand over responsibility of compiling ES from the ts plugin to babel, however. Currently typescript only compiles to ES2017, then bublé/babel does the rest. Any good reason it is done this way currently?

I guess the reasoning is to strip types with TS and do the actual transpilation to ES5 with buble/babel

src/index.js
```diff
@@ -226,6 +226,54 @@ export default async function microbundle(options) {
...
+
+	// Allow project to configure babel with local config file
+	if (hasBabelConfigFile) {
+		return {
+			extensions: [...DEFAULT_EXTENSIONS, '.ts', '.tsx'],
+		};
+	}
+
+	// No babelrc exists, use default config
+	// TODO move this into a separate babel-preset-microbundle package
```
FezVrasta: Could we start by moving this into its own file?

src/index.js
```diff
@@ -377,26 +425,7 @@ function createConfig(options, entry, format, writeMeta) {
...
-							],
-							[
-								require.resolve('@babel/plugin-proposal-class-properties'),
-								{ loose: true },
-							],
-						],
-					}),
+					babel(createBabelConfig(options)),
 					{
 						// Custom plugin that removes shebang from code because newer
```
FezVrasta: Do we still this? If yes maybe it's worth to update the comment to explain why

FezVrasta:
> Could you rebase on top of the master?

I sent a PR to @esphen's branch to rebase to master  
https://github.com/esphen/microbundle/pull/1

esphen:
New test results are in.

I've merged @FezVrasta's changes and updated the comment on the output sizes above. Notable changes:

- Added `async-ts` and `async-iife-ts` tests from master.
  - `async-ts` is down because babel has successfully tree-shaken the unused async function
  - `async-iife-ts` is flat
- Fixed the `esnext-ts` sizes, both before and after
  - `esnext-ts` is now up by 24.5%
  - Have not quite figured out why yet, feel free to dig in

Andarist:
@esphen have you updated the comparison table with new versions of those outputs? I might compare them later.

esphen:
@Andarist Yes, both the comparison table and the gist are up to date

wardpeet:
@ForsakenHarmony @developit do we want this getting merged? I can put my teeth in it as we want this for gatsbyjs too so we can add our own babelrc file or will we still keep babelrc false?

developit:
@wardpeet yeah. We definitely still want to move off Bublé.

PR #265, opened 2018-12-04
Remove import pruning

### LINKED ISSUE #264: mixed default and named exports not working in cjs and umd
```
const a=1, b=2, c=3, d=4;
export { a, b, c};
export default d;
```

if you try to bundle this, the result is:

```
module.exports=4;
//# sourceMappingURL=test3.js.map
```

I'd expect:

```
module.exports.a=1;
module.exports.b=2;
module.exports.c=3;
module.exports.default=4;
//# sourceMappingURL=test3.js.map
```

it exports everything only on the `esm` build.

Is there any flag to turn on?

marvinhagemeister:
Fixes #264 

**EDIT:** PR #262 likely needs to be merged first, as that one fixes the failing CI.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -12,7 +12,6 @@ import nodeResolve from 'rollup-plugin-node-resolve';
 import buble from 'rollup-plugin-buble';
 import { terser } from 'rollup-plugin-terser';
 import postcss from 'rollup-plugin-postcss';
-import alias from 'rollup-plugin-strict-alias';
 import gzipSize from 'gzip-size';
 import brotliSize from 'brotli-size';
 import prettyBytes from 'pretty-bytes';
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -291,19 +290,6 @@ function createConfig(options, entry, format, writeMeta) {
 	let nameCache = {};
 	let mangleOptions = options.pkg.mangle || false;
 
-	let exportType;
-	if (format !== 'es') {
```
developit: Could we make it an option instead of deleting?

marvinhagemeister: We could put this code behind an option. In either case it needs to be disabled for TypeScript because this breaks the generated typing definitions.

ForsakenHarmony:
bors try

ForsakenHarmony:
👀

ForsakenHarmony:
bors try

FezVrasta:
I just tested this out and it seems to work just fine. It'd be great to get this merged and released. It's pretty critical for the usage of the library.

developit:
I turned on branch protection so now we get the "update branch" button.

ForsakenHarmony:
bors retry

PR #268, opened 2018-12-10
feat: add support for Babel macros

FezVrasta:
I think it'd be very useful to support Babel macros out of the box, they allow to use Babel without additional configuration so they can improve a lot the developer experience without adding extra configuration to Microbundle.

Zero config tools such as create-react-app already support it.

### DIFF
--- a/test/fixtures/macro/src/macro.js
+++ b/test/fixtures/macro/src/macro.js
@@ -1,0 +1,20 @@
+import { createMacro } from 'babel-plugin-macros';
+
+function myMacro({ references, state, babel }) {
+	const { types: t } = babel;
+
+	references.default.forEach(referencePath => {
+		const parentPath = referencePath.findParent(t.isCallExpression);
+		let variableName = referencePath.findParent(t.isVariableDeclarator).node.id
+			.name;
+		if (
+			parentPath.node.arguments.length > 0 &&
+			parentPath.node.arguments[0] !== ''
+		) {
+			variableName = parentPath.node.arguments[0].value;
+		}
+		parentPath.replaceWith(t.stringLiteral(`${variableName}-macro`));
+	});
+}
+
+export const macro = createMacro(myMacro);
--- a/test/fixtures/macro/src/macro.js
+++ b/test/fixtures/macro/src/macro.js
@@ -3,7 +3,7 @@ import { createMacro } from 'babel-plugin-macros';
 function myMacro({ references, state, babel }) {
 	const { types: t } = babel;
 
-	references.default.forEach(referencePath => {
+	references.macro.forEach(referencePath => {
 		const parentPath = referencePath.findParent(t.isCallExpression);
 		let variableName = referencePath.findParent(t.isVariableDeclarator).node.id
 			.name;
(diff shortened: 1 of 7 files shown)


Andarist:
Personally I would wait on https://github.com/developit/microbundle/pull/263 before merging this to keep more natural evolution of the babel support in microbundle. OTOH some babel support is already merged in (async functions~) so there is no much harm in this one.

FezVrasta:
@Andarist can I do something to speed up the release cycle? At my company we recently started using microbundle for a big open source effort we are conducting that will lead to the public release of a lot of npm packages. We'd like to help as we can to take advantage of this library.

Thanks!

Andarist:
You can investigate why this fails - https://github.com/developit/microbundle/pull/263#issuecomment-444961022 . Or you can help @esphen implement this - https://github.com/rollup/rollup-plugin-babel/issues/275 (he has already started some work)

esphen:
@FezVrasta Thanks for offering to help out! 

I just dug into why the test on #263 fails, so no need to help out there anymore. Feel free to help out with rollup-plugin-babel if you want, though 🙂

This is a great suggestion, so we should definetely look into this once #263 is merged!

developit:
This seems useful enough to bring in.

Andarist:
This would require adding a test before merging.

FezVrasta:
may someone help me understand why the snapshots test are failing? The only difference seem the letters used to minify the variable names

test/fixtures/macro/src/macro.js
```diff
@@ -0,0 +1,20 @@
+const { createMacro } = require('babel-plugin-macros');
```
FezVrasta: this macro simply takes:

```
const name = macro();
```

and converts it to:
```
const name = 'name-macro';
```

FezVrasta:
Ok so, the issue was that the styled macro replaces its own import with `@emotion/styled-base`, and since that package isn't a dependency, it bundled it.

Adding it to the dependencies list is a "workaround", I'm not sure if there's any way to make this work in all the cases automatically tho.

FezVrasta:
@Andarist could we merge this now that Microbundle makes use of Babel internally?

ForsakenHarmony:
If you rebase once more 😇 I can merge it into dev which will be released soon (currently pre release)

FezVrasta:
Should I edit `/lib/babel-custom.js`?

ForsakenHarmony:
Probably 🙃

FezVrasta:
I rebased on top of `dev` but I now get this error in the new test this PR adds:

```
  ● fixtures › build macro with microbundle

    Cannot find module './macro' from '/Users/FezVrasta/Projects/microbundle/test/fixtures'
```

The import looks right, I'm not sure what's wrong

ForsakenHarmony:
Weird issue, seems to be a problem with babel plugin macros or something 🤔

or rollup babel

ForsakenHarmony:
nvm, our babel thing was broke

ForsakenHarmony:
oh, now everything else is broken 👌

ForsakenHarmony:
🎉🎉

PR #269, opened 2018-12-11
Fix async not transpiled with TS

### LINKED ISSUE #259: TypeScript - Unexpected token v0.8.3
example file
```ts
export class MyClass {
    async foo() {}
}
```

error
```
> microbundle build --target node --format cjs,es -o dist

Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)

at /index.ts:2:10

1: export class MyClass {
2:     async foo() {}
             ^
3: }
```

repo: https://github.com/scttcper/microbundleunexpected

marvinhagemeister:
Fixes #259.

### DIFF
--- a/test/__snapshots__/index.test.js.snap
+++ b/test/__snapshots__/index.test.js.snap
@@ -1,5 +1,63 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`fixtures async-iife-ts 1`] = `
+"Used script: microbundle
+
+Directory tree:
+
+async-iife-ts
+  dist
+    async-iife-ts.js
+    async-iife-ts.js.map
+    async-iife-ts.mjs
+    async-iife-ts.mjs.map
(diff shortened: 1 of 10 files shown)


test/__snapshots__/index.test.js.snap
```diff
@@ -310,12 +368,12 @@ esnext-ts
 
 
 Build \\"esnextTs\\" to dist:
-470 B: esnext-ts.js.gz
-404 B: esnext-ts.js.br
-470 B: esnext-ts.mjs.gz
-409 B: esnext-ts.mjs.br
-528 B: esnext-ts.umd.js.gz
-461 B: esnext-ts.umd.js.br"
+841 B: esnext-ts.js.gz
```
marvinhagemeister: While the size increase is looking substantial, we were previously not transpiling all `asnyc/await` keywords in the example. This PR fixes that

PR #272, opened 2018-12-13
Bugfix: still enable Terser optimizations when eval is used

developit:
This fixes a bug where Microbundle's output is suboptimal if `eval()` or `new Function()` is used.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -587,7 +587,7 @@ function createConfig(options, entry, format, writeMeta) {
 							warnings: true,
 							ecma: 5,
 							toplevel: format === 'cjs' || format === 'es',
-							mangle: Object.assign({}, minifyOptions.mangle || {}),
+							mangle: Object.assign({ eval: true }, minifyOptions.mangle || {}),
 							nameCache,
 						}),
 						nameCache && {


rpetrich:
Does this emit a warning when `eval` or `new Function` is used? Mangling when `eval` is used changes observable behaviour since code passed to `eval` can reference bindings in any of the parent scopes. Mangling parents of `new Function` should be safe.

developit:
@rpetrich Not sure. My use-case is actually `new Function`, but I can't seem to find a Terser option to enable mangling only for that case.

I believe the reason this is disabled by default is because Terser does not (cannot) mangle the string passed to eval/Function, so it's possible that string contains references to variables in the outer scope that would have changed names.

This unfortunately often has the consequence of "de-opting" an entire module since its all in scope of the eval/Function.

We could potentially add our own warning. Folks shouldn't be using eval in the first place, but there are some extremely specific cases where it can be used that shouldn't come with this side-effect.

rpetrich:
Interesting. I agree that `eval` should basically never be used as it forces a deopt on all ancestor scopes, but my understanding is that `new Function`/`eval` called through a variable doesn't. It's a shame that Terser doesn't split these into separate options.

developit:
@rpetrich another option here would be to extend the existing `.mangle` `package.json` field with an `eval:true` option.  I'd prefer not to since its something nobody will ever do, but if its the only way to get eval in then it'd be worthwhile.

PR #273, opened 2018-12-14
Slightly improve Terser optimizations config

developit:
This reduces the size of a certain well-known library from 3130b to 3104b.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -578,6 +578,8 @@ function createConfig(options, entry, format, writeMeta) {
 							},
 							compress: Object.assign(
 								{
+									hoist_funs: true,
+									booleans_as_integers: true,
 									keep_infinity: true,
 									pure_getters: true,
 									passes: 10,
@@ -586,6 +588,7 @@ function createConfig(options, entry, format, writeMeta) {
 							),
 							warnings: true,
 							ecma: 5,
+							module: format === 'es',
 							toplevel: format === 'cjs' || format === 'es',
 							mangle: Object.assign({}, minifyOptions.mangle || {}),
 							nameCache,
--- a/src/index.js
+++ b/src/index.js
@@ -579,7 +579,6 @@ function createConfig(options, entry, format, writeMeta) {
 							compress: Object.assign(
 								{
 									hoist_funs: true,
-									booleans_as_integers: true,
 									keep_infinity: true,
 									pure_getters: true,
 									passes: 10,


src/index.js
```diff
@@ -449,11 +449,15 @@ function createConfig(options, entry, format, writeMeta) {
 							sourcemap: true,
 							output: { comments: false },
 							compress: {
+								passes: 10,
+								hoist_funs: true,
+								booleans_as_integers: true,
```
Andarist: Based on its description this seems like really unsafe

src/index.js
```diff
@@ -449,11 +449,15 @@ function createConfig(options, entry, format, writeMeta) {
 							sourcemap: true,
 							output: { comments: false },
 							compress: {
+								passes: 10,
+								hoist_funs: true,
```
Andarist: hm, funny that it makes sense - would assume that this might make things worse, unhoisted should have generally speaking shorter identifiers than hoisted ones

src/index.js
developit: removed it and added an escape hatch so folks can specify `compress:{}` in the existing package.json mangle configuration.

src/index.js
```diff
@@ -448,12 +448,15 @@ function createConfig(options, entry, format, writeMeta) {
...
 							sourcemap: true,
 							output: { comments: false },
-							compress: {
+							compress: Object.assign({
+								passes: 10,
+								hoist_funs: true,
 								keep_infinity: true,
 								pure_getters: true,
-							},
+							}, mangleOptions && mangleOptions.compress || {}),
```
Andarist: IMHO this is confusing for uglify/terser users - because those are separate options there & not nested one like here.

src/index.js
marvinhagemeister: I can vouch for the performance improvements 👍

src/index.js
marvinhagemeister: Perhaps we could use a `terser.config.json` file. It could look something like this:

```json
{
  "mangle": { ... },
  "compress": { ... }
}
```

Andarist: Well - IMHO we should support `microbundle.config.js` (or similar), personally I hate putting tool configs in package.json (but I understand some people like it), I mean tool-specific config, I'm on board with using some standard package.json fields as base for the tool config as convention. To support both we could just use https://github.com/davidtheclark/cosmiconfig

Im opposed to adding `terser.config.js` because it would suggest to users that Terser itself supports config files, which it doesnt

I plan to start a big refactoring of `microbundle` (in parts, so it would be easy to discuss, follow and review), maybe this could just sit until I do it? Otherwise we might release a feature which potentially might be changed in the near future.

marvinhagemeister: That's good to know. What are the main things you want to tackle with the refactor? As I'm new to `microbundle` I'm not up to speed on current plans.

Andarist: What I would like to tackle:
- config files
- enforce convention (need consensus on this one) with autofixer, instead of having defaults for some things (like i.e. main/module fields), just force their output filenames, but provide autofixing feature so developer can knowingly accept the changes
- validate package.json shape and certain code constructs (i.e. when detecting `typeof window`/`typeof document` enforce people to include `"browser"` field in package.json (again with autofix feature), ofc build browser bundles with those values inlined (smaller bundles 🚀)
- monorepo support
- more react-specific optimizations builtin
- when we land babel@7 support - use `@babel/runtime` helpers and offer to install that dep with autofix
- opt-in react-native bundles
- multiple entry points support (like `redux-saga` & `redux-saga/effects`) with rollup's code-splitting

Probably more 😅 plan to work on it slowly but steadily.

marvinhagemeister: Awesome, thanks for sharing! That's quite a roadmap and we should see how much we can split the tasks, so we keep being able to make timely releases 👍

I'm not sure how much time I have to contribute to those, but if I can I'd be happy to help out.

Regarding config files I'm not sure how much we want to allow users to customize besides `terser`, `babel` and `TS/Flow` options. Only the `terser` doesn't support config files on its own whereas all the others do support it already. We just have blocked it for `babel` for now until the switch away from `bublé` is complete. Perhaps we can discuss this upstream with the `terser` maintainers.

@developit Do we still need more custom `compress` settings for preact x or are the defaults you changed here good enough?

Andarist: I dont want too many things to be customizable, we have some options already though - https://github.com/developit/microbundle#all-cli-options and at least those should be supported. Also I would like to be able to override some of those options per bundle type.

developit:
I think we can close this? I added support in another PR.
(example: https://github.com/developit/preact/blob/master/mangle.json#L6-L16)

PR #279, opened 2018-12-18
Remove import pruning

ForsakenHarmony:


marvinhagemeister:
What's different from #265 ?

ForsakenHarmony:
Oh, I forgot about that :^)

just had another problem with `index.ts.js` and created this without thinking

PR #287, opened 2018-12-23
0.9.0

ForsakenHarmony:


marvinhagemeister:
Merging, as the changes in this PR are harmless. We need to check out why there is a failure though. My guess is that this is caused by transient dependency update, but that's just a guess for now.

ForsakenHarmony:
bors try

ForsakenHarmony:
bors r=marvinhagemeister

PR #290, opened 2018-12-25
Refactor | Test refactoring

MatiasOlivera:
Make the code clearer and readable


test/index.test.js
```diff
@@ -14,23 +14,18 @@ const rimraf = promisify(_rimraf);
...
-				node =>
-					`${indent}${node.name}\n${
-						node.type === 'directory'
-							? printTree(node.children, indentLevel + 1)
-							: ''
-					}`,
-			),
+			.map(node => {
+				const isDir = node.type === 'directory';
+				const dirTree = isDir ? printTree(node.children, indentLevel + 1) : '';
```
Andarist: imho this would be better inlined, after all its not a `dirTree` in `!isDir` case

test/index.test.js
```diff
@@ -60,13 +55,24 @@ describe('fixtures', () => {
...
-				));
-			} catch (err) {}
-			script = script || DEFAULT_SCRIPT;
+			async function getBuildScript() {
+				try {
+					const pkgJSON = await readFile(
+						resolve(fixturePath, 'package.json'),
+						'utf8',
+					);
+					const pgk = JSON.parse(pkgJSON);
```
Andarist: typo, should be `pkg`

test/index.test.js
```diff
@@ -60,13 +55,24 @@ describe('fixtures', () => {
...
-			let script;
-			try {
-				({ scripts: { build: script } = {} } = JSON.parse(
-					await readFile(resolve(fixturePath, 'package.json'), 'utf8'),
-				));
-			} catch (err) {}
-			script = script || DEFAULT_SCRIPT;
+			async function getBuildScript() {
+				try {
+					const pkgJSON = await readFile(
```
Andarist: actually would be better to check if this file exists before reading it unconditionally - that way lack of package.json could be handled more explicitly here

test/index.test.js
```diff
@@ -60,13 +55,24 @@ describe('fixtures', () => {
...
 			await rimraf(resolve(`${fixturePath}/.rts2_cache_umd`));
 
-			let script;
-			try {
-				({ scripts: { build: script } = {} } = JSON.parse(
-					await readFile(resolve(fixturePath, 'package.json'), 'utf8'),
-				));
-			} catch (err) {}
-			script = script || DEFAULT_SCRIPT;
+			async function getBuildScript() {
```
Andarist: personally i would prefer adding a parameter to this and hoisting it to the top level scope

test/index.test.js
```diff
@@ -60,13 +55,24 @@ describe('fixtures', () => {
...
+					return build;
+				} catch (err) {
+					// Pass the no-pkg test
+					// because it does not have the package.json file
+					return DEFAULT_SCRIPT;
+				}
+			}
+
+			const build = await getBuildScript();
+			const script = build || DEFAULT_SCRIPT;
```
Andarist: defaulting should be handled by `getBuildScript`

test/index.test.js
MatiasOlivera: You are right, I will change it.

test/index.test.js
MatiasOlivera: Yes, that's better

test/index.test.js
MatiasOlivera: That makes sense to me

test/index.test.js
```diff
@@ -14,26 +14,46 @@ const rimraf = promisify(_rimraf);
...
+			return defaultScript;
+		}
+	}
+
+	try {
+		const pkgJSON = await readFile(pkgJsonPath, 'utf8');
+		const pkg = JSON.parse(pkgJSON);
+		const { scripts: { build } = {} } = pkg;
+
+		// Return default script if no have
```
Andarist: imho this comment is excessive, what `return build || defaultScript` does is pretty obvious without it

test/index.test.js
```diff
@@ -14,26 +14,46 @@ const rimraf = promisify(_rimraf);
...
+
+	try {
+		await stat(pkgJsonPath);
+	} catch (err) {
+		if (err.code === 'ENOENT') {
+			return defaultScript;
+		}
+	}
+
+	try {
```
Andarist: is this try/catch needed now?

test/index.test.js
```diff
@@ -14,26 +14,46 @@ const rimraf = promisify(_rimraf);
...
+				}`;
+			}),
 	);
 };
 
+const getBuildScript = async (fixturePath, defaultScript) => {
+	const pkgJsonPath = resolve(fixturePath, 'package.json');
+
+	try {
+		await stat(pkgJsonPath);
```
Andarist: according to - https://nodejs.org/api/fs.html#fs_fs_stat_path_options_callback 
> Using fs.stat() to check for the existence of a file before calling fs.open(), fs.readFile() or fs.writeFile() is not recommended. Instead, user code should open/read/write the file directly and handle the error raised if the file is not available.

MatiasOlivera: Yes, I had also read it. What do we do?

test/index.test.js
MatiasOlivera: Yes, I don't know why I added it

test/index.test.js
MatiasOlivera: IMHO I think it makes the intention clear, anyway I am open to suggestions to improve it

test/index.test.js
developit:
```js
let pkg = {};
try {
  pkg = JSON.parse(await readFile(pkgJsonPath, 'utf8'));
} catch (e) {}
return pkg && pkg.scripts && pkg.scripts.build;

File: test/index.test.js [hunk 9]
```diff
@@ -14,26 +14,32 @@ const rimraf = promisify(_rimraf);
...
 	);
 };
 
+const getBuildScript = async (fixturePath, defaultScript) => {
+	let pkg = {};
+	try {
+		pkg = JSON.parse(
+			await readFile(resolve(fixturePath, 'package.json'), 'utf8'),
+		);
+	} catch (e) {}
```
Andarist: shouldnt this check `e.code` and rethrow if its not the error we are looking for?

ForsakenHarmony: this wasn't done before either

Andarist: yeah, sure - but this PR is an opportunity to improve the code here, so IMHO it could be improved as part of it

MatiasOlivera: I think it could save debugging time in the future if there are problems with some package.json

Issue #288, opened 2018-12-24
Add Rollup plugin with support for Node built-in modules


gaui:
When importing `child_process` and targeting node, I still get this error from #238 :

```
'child_process' is imported by src\command.ts, but could not be resolved – treating it as an external dependency
Error: Could not load child_process (imported by src/command.ts): ENOENT: no such file or directory, open 'child_process'
```

This can be bypassed by adding `--external child_process` argument. Still I get this message:

> Creating a browser bundle that depends on Node.js built-in modules ('path', 'assert', 'util' and 'events'). You might need to include https://www.npmjs.com/package/rollup-plugin-node-builtins

So please add package `rollup-plugin-node-builtins` so Rollup knows how to detect Node built-in modules and resolve them appropriately.

marvinhagemeister:
Good catch, can you make a PR?

lukeed:
I think we can skip the plugin & do a special case for `--external natives` that adds them:

```js
require('module').builtinModules || Object.keys(process.binding('natives'))
```

gaui:
@lukeed Spot on. But wouldn't that resolution logic belong in Rollup?

lukeed:
@gaui Personally, I would vote yes, but realistically it's not something Rollup can/wants to do because you may have shims/custom modules that share the same names, and the above would likely (have to) override anything non-native with the native version.

gaui:
If you're naming your modules the same as the natives, you're doing something wrong. 😄

gaui:
@Rich-Harris what do you think?

gaui:
@lukeed Is it possible to mark those modules returned from the above function, as external so Rollup knows? Not familiar with microbundle internals.

z0al:
@lukeed 
> I think we can skip the plugin & do a special case for `--external natives` that adds them

I don't get it. How this would add necessary shims when, for example, bundling targeting the browser?

PR #303, opened 2019-01-18
Add ability to include node built-in modules as external dependencies

krismuniz:
Hello 👋

This PR adds the ability to include node built-in modules as external dependencies by specifying `--external natives`

> Related to https://github.com/developit/microbundle/issues/288, but I'm not 100% sure if it closes it, as the PR suggests using the `rollup-plugin-node-builtins` plugin for resolving these dependencies and including shims where necessary (i.e. in the browser).

### Problem

When using `microbundle` for building node modules (`--target node`), you have to include every built-in module like this:

```shell
microbundle --external events,util,child_process
```

This becomes even more tedious when a built-in module uses _other_ built-in modules.

As @lukeed suggests in [this comment](https://github.com/developit/microbundle/issues/288#issuecomment-449825403), it would be nice to have the option to simply treat all built-in node modules as external dependencies by adding `--external natives` instead of `--external events,util,child_process[, ...]`. 

### Solution

Add all modules from `require('module').builtinModules` to the list of external dependencies like this:

```shell
microbundle --external natives
```

### Feedback Welcome

 ✏️Let me know if you need any changes
🗯If this is not ideal, what would be a nice alternative to specifying all built-in modules?


lukeed:
> Something like this, right?

Correct!

The only other thing to consider here is that `require('module').builtinModules` doesn't work in Node 4.x – hence my two-part suggestion in the original issue. I don't know/can't remember the minimum supported Node version `microbundle` is targeting.

Andarist:
I also agree that this should be the default with node as target, we could use this static list for this - https://github.com/sindresorhus/builtin-modules/blob/master/builtin-modules.json

lukeed:
Don't think we need an extra dependency – the list is built-in to Node in all versions:

```js
require('module').builtinModules || Object.keys(process.binding('natives'))
```

krismuniz:
Right! So if we want Node 4.x support we'd do something like this:

```diff
- let external = require('module').builtinModules.concat(/* ... */);
+ const builtinModules = require('module').builtinModules || Object.keys(process.binding('natives'));
/* ... */
+ builtinModules.concat(/* ... */)
```

developit:
@krismuniz it should only apply to `target==='node'` yup.

ianstormtaylor:
Any news on this? This would be super helpful.

ForsakenHarmony:
Problems mentioned above would have to be solved

developit:
FWIW this behavior is already enabled by default in `microbundle@next`, since `rollup-plugin-node-resolve` handles it by default. I've opened #591 to silence the warnings.

PR #305, opened 2019-01-20
add mdx-deck-live-code to "Built with..." section

JReinhold:
Microbundle is JUST the bundling tool I've been looking for, thanks for this!

I would be honoured to have my library mentioned in the "Built with Microbundle" section.

repo: https://github.com/JReinhold/mdx-deck-live-code
demo: https://reinhold.is/live-coding-in-slides


developit:
I might move these into a wiki if the list keeps growing, and then the readme would just be microbundle-powered tools like microenvi

PR #310, opened 2019-01-29
Fixed ES Modules bundle path

slorenzo:
Ey @developit. I have fixed the ES bundle path on the README.md file. It could cause some issues if the devs copy and paste.


README.md
```diff
@@ -85,7 +85,7 @@ You can specify output builds in a `package.json` as follows:
 ```
 "main": "dist/foo.js",          // CJS bundle
 "umd:main": "dist/foo.umd.js",  // UMD bundle
-"module": "dist/foo.mjs",      // ES Modules bundle
+"module": "dist/foo.m.js",      // ES Modules bundle
```
JoviDeCroock: mjs is allowed, it can be read with webpack loader `javascript/auto` but another alternative good name could be dist/foo.module.js but foo.m.js seems quite weird.

ForsakenHarmony:
what is this supposed to fix?

ForsakenHarmony:
https://github.com/developit/microbundle/pull/213

slorenzo:
Maybe I'm wrong, but I do not see any file created with the extension `mjs`. Instead, a file with `.m.js` extension was created.

developit:
@slorenzo hmm, we changed from `.m.js` to `.mjs` in [Microbundle 0.7.0](https://github.com/developit/microbundle/releases/tag/0.7.0), any chance you're using an older version?

slorenzo:
@developit I was wrong and I have fixed it in my project. Anyway, some packages are doing the same.

Take a look here:
https://github.com/philipp-spiess/react-recomponent/blob/master/package.json#L7
https://github.com/tmcw/wcag-contrast/blob/master/package.json#L9

It's a quite confused.

developit:
hmm - I think that's okay though?  As long as we use that name unmodified its fine if folks set it to something other than `.mjs`.

ForsakenHarmony:
Well webpack was somewhat broken so might as well revert this

PR #311, opened 2019-01-30
Don't override tsconfig (fixes #300)

### LINKED ISSUE #300: 0.0.7+ TypeScript target options from tsconfig.json ignored
This seems to be the offending line in the source: https://github.com/developit/microbundle/blob/master/src/index.js#L411
and the commit: https://github.com/developit/microbundle/commit/cebff34d415b00d424876f25f1a700fd4b17e21c

I assume, based on the commit, that this was added since typescript is compiled before additional transforms are added.

I'm not quite sure what direction this library is headed in, so I'd love some feedback. Would it make sense to have the `tsconfig.json` of the user be the source of truth for setting the target output or have it configurable via microbundle itself? I assumed that the `--target` flag could be used, but it only appears to modify the module resolution.

Either way, I think microbundle can stand to have the output be configurable.

Cheers!

developit:
See https://github.com/ezolenko/rollup-plugin-typescript2/blob/master/src/ioptions.ts#L27

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -408,7 +408,7 @@ function createConfig(options, entry, format, writeMeta) {
 									jsx: options.jsx,
 								},
 							},
-							tsconfigOverride: {
+							tsconfigDefaults: {
 								compilerOptions: {
 									target: 'esnext',
 								},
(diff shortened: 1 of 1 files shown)


developit:
ah true.

PR #312, opened 2019-01-30
Add define option

developit:
Same as Terser's `--define`:

```
microbundle --define NODE_ENV=production
# this is also now set as the default ^
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -494,7 +494,7 @@ function createConfig(options, entry, format, writeMeta) {
 							compress: {
 								keep_infinity: true,
 								pure_getters: true,
-								global_defs: defines
+								global_defs: defines,
 							},
 							warnings: true,
 							ecma: 5,
--- a/src/index.js
+++ b/src/index.js
@@ -488,6 +488,10 @@ function createConfig(options, entry, format, writeMeta) {
 							compress: {
 								keep_infinity: true,
 								pure_getters: true,
+								global_defs: Object.assign({
+									NODE_ENV: process.env.NODE_ENV || 'production',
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -311,7 +312,16 @@ function createConfig(options, entry, format, writeMeta) {
 		return globals;
 	}, {});
 	if (options.globals && options.globals !== 'none') {
-		globals = Object.assign(globals, parseGlobals(options.globals));
+		globals = Object.assign(globals, parseMappingArgument(options.globals));
+	}
+
+	const NODE_ENV = process.env.NODE_ENV || 'production';
```
ForsakenHarmony: this is for libraries

ForsakenHarmony: so I don't support having that as the default at this point in the "toolchain"

jviide: That is a good point. `NODE_ENV` and `process.env.NODE_ENV` are probably best left undefined altogether by default.

developit: @ForsakenHarmony think it's better to not do inlining by default?  I'm good with that.

developit:
Also apparently I'd managed to miss #233 when I opened this.

weslleyaraujo:
Is there any plans for a release including this soon? 😄
that would be super helpful on the project I am working on 👍

guicamillo:
> Is there any plans for a release including this soon? 😄
> that would be super helpful on the project I am working on 👍

Same thing over here :D

developit:
yup

ChrisBrownie55:
Since `--define` was released in version `0.10.1` it should probably be added to the README, seems like the only place it's documented currently is the releases tab.

PR #317, opened 2019-02-15
chore: documented the `--no-compress` option

mweststrate:


src/prog.js
```diff
@@ -24,7 +24,8 @@ export default handler => {
 		.option('--external', `Specify external dependencies, or 'none'`)
 		.option('--globals', `Specify globals dependencies, or 'none'`)
 		.example('microbundle --globals react=React,jquery=$')
-		.option('--compress', 'Compress output using Terser', null)
+		.option('--compress', 'Compress output using Terser (default)', null)
+	        .option('--no-compress', 'Disable output compressing', null)
```
ForsakenHarmony: sade automatically adds that option, but yes it's confusion

ForsakenHarmony: and ci is complaining about the indent here 🙃

mweststrate: yeah, took me diving in the source to figure how to disable compress, as `--compress=false` and similar doesn't work!

mweststrate: Ok, attempted to fix, hope it worked out (creating this PR from the GitHub interface)

ForsakenHarmony: Still want to make it the default, maybe soon

ForsakenHarmony:
Have you tested that adding the option doesn't break the behaviour?

mweststrate:
No

Op vr 15 feb. 2019 15:49 schreef Leah <notifications@github.com>:

developit:
Idea: add a new option `--uncompressed` and alias `--compress` to it, then only document `--uncompressed`.

maraisr:
Here's a silly thought, how about we remove the `--compress`. Assume you always wan't to compress for builds, and merely _disable_ it with a `--no-compress`. Would dramatically simplify the api, and solves the whole `--compress --no-compress` scenario, which one wins out.

developit:
I'm fine merging this if it doesn't mess up arg parsing. The documentation at least should be merged.

PR #321, opened 2019-02-22
Set the number of terser passes to 10

jviide:
This pull request (based on a conversation with @developit) sets Terser's `passes` argument to `10`. The number of passes is the "maximum number of times to run compress", and may help in some cases like when using the new `--define` parameter.


developit:
magic

developit:
image

Issue #304, opened 2019-01-20
RFC: option for modern bundling.


JoviDeCroock:
It's 2019 and browsers have caught up, 85% of browsers support classes and async/await. This makes me think that transpiling down to es5 won't benefit everyone since not everyone has to support older browsers.

Historically the responsibility as to how far transpiled down a bundle was has always been the choiche of the author.
I wanted to "feel the water" in something also suggested by Henry Zhu (author of babel) where authors would publish their `module` field as an ES6/ES2015 package and give the option to the developer to transpile this down if the need  for IE11/... is needed.

In essence this shouldn't pose an issue, adding a package to your `babel-loader` or `tsconfig`.
I do think that it would be the responsibility of _microbundle_ to document this option well enough so library authors are made aware of the possible implications for their consumers.

I also made a small POC showing the difference in [size](https://github.com/JoviDeCroock/POC-ModularLegacyBuild). this is a shippable build that will fallback to a legacy bundle when need be.

So in essence, the developer working with this could choose to transpile down library x and y for their legacy bundle and not for the modern one.

To sum up and clarify the whole point, I would not say all module builds should be es6. I would suggest an option for the library author to build a more modern bundle.

Thank you all for reading!

Sources
[class](https://caniuse.com/#search=class)
[async](https://caniuse.com/#search=async)
[Deploying ES6](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/)
[babel ES6](https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages)

ForsakenHarmony:
the only problem with it is how to specify which file is the modern bundle

JoviDeCroock:
> the only problem with it is how to specify which file is the modern bundle

It would be the module field by default, no es5 transpiled file only the es6 one.
Maybe I don't fully grasp what you are implying

Andarist:
The problem is that most of the tools, tutorials in the wild etc suggest that `node_modules` should not be transpiled by consumers, that main/module should contain already ready-to-be-consumed es5. 

There are 2 ways of handling this:
- go against current expectations and start shipping es6+ in main/module
- create a new field that could be an opt-in for bundlers understanding it

I guess @ForsakenHarmony was referring to the latter.

JoviDeCroock:
I completely agree with that statement but I don't really know if we should keep adding things to package.json. Most people don't even know what all of it means.

[...]

PR #335, opened 2019-03-04
Add compressor options

developit:
This allows specifying Terser options in mangle.json or package.json. The old `"mangle"` field is still supported for legacy reasons alongside the new `"minify"` option.

```js
// in `mangle.json` or `package.json`:
{
  "minify": {
    // old/simple variant:
    "regex": "^_",
    // new: customize any terser options:
    "compress": {
      "hoist_vars": true,
      "reduce_funcs": false
    }
  }
}
```

This is useful for projects that are already structured with functions aimed at gzip compression optimization.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -400,6 +400,9 @@ function createConfig(options, entry, format, writeMeta) {
 			nameCache = JSON.parse(
 				fs.readFileSync(resolve(options.cwd, 'mangle.json'), 'utf8'),
 			);
+			if (nameCache.config) {
+				mangleOptions = Object.assign({}, mangleOptions || {}, nameCache.config);
+			}
 		} catch (e) {}
 	}
 	loadNameCache();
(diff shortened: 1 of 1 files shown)


developit:
I'd like to propose an alternative:

```js
// via mangle.json
{
  // options map roughly to https://github.com/terser-js/terser#minify-options-structure
  "minify": {
    "properties": "^_",  // current easy way to set property mangling regexp

    "compress": {  // optional
      "hoist_vars": true,
      "reduce_funcs": false
    },

    "mangle": {}  // optional
  }
}
```

developit:
@marvinhagemeister for now I've left out a couple top-level configuration items like `module` and `ecma` because Microbundle is going to automate those soon and we don't want folks to tweak them thinking it's necessary.

developit:
@hassanbazzi most of the Terser options folks would want to change are on `mangle` and `compress`, so for those that's what this does - sets some defaults then allows arbitrary overrides.

For Terser's root options, we really can't allow arbitrary customization. Doing so would break output
in the very near future, when Microbundle begins outputting ES2017 alongside the current downlevelled ES5 (see #304). When we do this, we'll be running Terser in ES5 mode for the 3 classic script bundles, and `{ecma:8,module:true}` for modern bundles.  Since `ecma` and `module` are the two main root options we're not exposing here, I think that with the modern bundling setup in place there won't really be a reason to customize any root options. We can always add them manually once we have a better idea, but for now I don't want folks to try customizing `ecma` only to intentionally break it in 1.0.

Issue #331, opened 2019-02-28
typescript definitions directory


manferlo81:
I just found out about **microbundle**, I tried it and I think it's awesome, however I cant manage to set the declaration files directory other that the **dist** folder. I tried setting the "**types**" and "**typings**" fields in the **package.json** file, I also tried to set the "**declarationDir**" of my **tsconfig.json** file without success.

I am starting a project using **rollup** but now I am considering to switch to **microbundle**, it will save me some config time.

I've noticed the "**useTsconfigDeclarationDir**" option in the **rollup-plugin-typescript2** has not been set to **true**, did that cause any issues in the past? otherwise it would be a good idea to set it. If you wish I'd be more than happy to create a pull request.

Alternatively a **CLI** option can be setup (_--declaration_ or _--declartionDir_ for example) or get the directory from the "**types**" or "**typings**" of the **package.json**.

developit:
Ah interesting - are you saying that if we had Microbundle use `useTsconfigDeclarationDir:true`, it woudl respect the `typings` field you defined in your package.json and output d.ts files there instead?  If so that sounds like a great feature.

manferlo81:
> Ah interesting - are you saying that if we had Microbundle use `useTsconfigDeclarationDir:true`, it woudl respect the `typings` field you defined in your package.json and output d.ts files there instead? If so that sounds like a great feature.
> 
> In general, we're vastly in favor of infering any configuration from well-known package.json fields whenever possible. This seems like a great case where we can!

In fact useTsconfigDeclarationDir: true will respect the "declarationDir" in your tsconfig.json file, but it would take very little effort to implement for it to check the "typings" (or "types") fields in your package.json as a fallback, in case you don't really need a tsconfig.json file.

I cloned your repo yesterday,  I will try later today and I will edit this comment with my ideas.

manferlo81:
Hi there, I think you didn't get notified about my [previous comment](https://github.com/developit/microbundle/issues/331#issuecomment-468884481), Anyway  I am starting my new library using **rollup** because I need my types in a different directory, looking forward to use **microbundle** next time, it would really make my life easier. I hope my ideas and comments serve you in some way. Need some help, just ask, happy to help. Once again great job!.

marvinhagemeister:
Should be easy enough to add 👍

Glitches:
I'm trying to write this, and if it goes well a PR will follow

PR #351, opened 2019-03-14
Feat/ts definition directory

Glitches:
Added custom types build path based upon the Issue: [](https://github.com/developit/microbundle/issues/331)


ForsakenHarmony:
mind removing the tar file and rebasing?

Glitches:
> mind removing the tar file and rebasing?

Done!

wardpeet:
It seems to be already fixed by https://github.com/developit/microbundle/pull/646. It doesn't have an extra declarationDir option, but it does check types & typings keys in package.json, which should be sufficient.

PR #354, opened 2019-03-16
Readme/newcomer infos

isthatcentered:
Hello,

I discovered the library today and found it really amazing. As a newcomer, I found a few informations were missing, so here is my proposal.

This makes the default input/output paths behavior a bit more easier to find and adds typescript to the list of features.

Thank you very much for the amazing work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Libraries often wish to rename internal object properties or class members to sm
 ### All CLI Options
 
 ```
-  Usage
+ Usage
     $ microbundle <command> [options]
 
   Available Commands
@@ -101,16 +101,25 @@ Libraries often wish to rename internal object properties or class members to sm
     -o, --output     Directory to place build files into
     -f, --format     Only build specified formats  (default es,cjs,umd)
     -w, --watch      Rebuilds on any change  (default false)
-    --target         Specify your target environment (node or web) (default web)
+    --target         Specify your target environment (node or web, default web)
     --external       Specify external dependencies, or 'none'
     --globals        Specify globals dependencies, or 'none'
+    --define         Replace constants with hard-coded values
+    --alias          Map imports to different modules
     --compress       Compress output using Terser  (default true)
     --strict         Enforce undefined global context and add "use strict"
     --name           Specify name exposed in UMD builds
     --cwd            Use an alternative working directory  (default .)
     --sourcemap      Generate source map  (default true)
-    -h, --help       Displays this message
+    --raw            Show raw byte size  (default false)
     --jsx            A custom JSX pragma like React.createElement (default: h)
+    -h, --help       Displays this message
+
+  Examples
+    $ microbundle microbundle --globals react=React,jquery=$
+    $ microbundle microbundle --define API_KEY=1234
+    $ microbundle microbundle --alias react=preact
+    $ microbundle microbundle --no-sourcemap # don't generate sourcemaps
 ```
 
 ## 🛣 Roadmap
--- a/README.md
+++ b/README.md
@@ -63,8 +63,9 @@ You can specify output builds in a `package.json` as follows:
 ```
 "main": "dist/foo.js",          // CJS bundle
 "umd:main": "dist/foo.umd.js",  // UMD bundle
-"module": "dist/foo.mjs",      // ES Modules bundle
+"module": "dist/foo.mjs",       // ES Modules bundle
 "source": "src/foo.js",         // custom entry module (same as 1st arg to microbundle)
+"types": "dist/foo.d.ts",       // Typescript typings
 ```
 
 ### Mangling Properties
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 - Produces tiny, optimized code for all inputs
 - Supports multiple entry modules _(`cli.js` + `index.js`, etc)_
 - Creates multiple output formats for each entry _(<abbr title="CommonJS (node)">CJS</abbr>, <abbr title="Universal Module Definition">UMD</abbr> & <abbr title="ECMAScript Modules">ESM</abbr>)_
-- 0 configuration Typecript support
+- 0 configuration TypeScript support
 - Built-in Terser compression & gzipped bundle size tracking
 
 ## 🔧 Installation
@@ -30,7 +30,7 @@
 ```js
 {
   "source": "src/foo.js",         // Your source file (same as 1st arg to microbundle)
-  "main": "dist/foo.js",          // Desire output path
+  "main": "dist/foo.js",          // Desired output path
   "scripts": {
     "build": "microbundle",       // uses the "source" and "main" key in the file as input and output paths by default
     "dev": "microbundle watch"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,13 +44,41 @@ Microbundle includes two commands - `build` (the default) and `watch`. Neither r
 
 ### `microbundle` / `microbundle build`
 
-By default, microbundle will infer the location of your source entry file
-(the root module in your program) from the `source` field in your `package.json`. It will infer the output directory and filename(s) from the `main` field. For UMD builds, microbundle will use a snake case version of the `name` field in your `package.json` for the export name; you can also specify a name via an `amdName` field or the `name` CLI option.
+Unless overridden via command line, microbundle uses the `source` property in your `package.json` to locate the input file, and the `main` property for the output.
```
ChrisBrownie55:
```suggestion
Unless overridden via the command line, microbundle uses the `source` property in your `package.json` to locate the input file, and the `main` property for the output.
```

README.md
```diff
@@ -16,18 +16,23 @@
...
+
 `npm i -D microbundle`
 
-... then add the scripts to your `package.json`:
+### Set up your `package.json`
 
 ```js
 {
+  "source": "src/foo.js",         // Your source file (same as 1st arg to microbundle)
+  "main": "dist/foo.js",          // Desire output path
```
ChrisBrownie55:
```suggestion
  "main": "dist/foo.js",          // Desired output path
```

README.md
```diff
@@ -39,13 +44,41 @@ Microbundle includes two commands - `build` (the default) and `watch`. Neither r
...
 
-By default, microbundle will infer the location of your source entry file
-(the root module in your program) from the `source` field in your `package.json`. It will infer the output directory and filename(s) from the `main` field. For UMD builds, microbundle will use a snake case version of the `name` field in your `package.json` for the export name; you can also specify a name via an `amdName` field or the `name` CLI option.
+Unless overridden via command line, microbundle uses the `source` property in your `package.json` to locate the input file, and the `main` property for the output.
+
+For UMD builds, microbundle will use a snake case version of the `name` field in your `package.json` as export name. This can be overridden either by providing an `amdName` key in your `package.json` or via the `--name` flag in the cli.
 
 ### `microbundle watch`
 
 Just like `microbundle build`, but watches your source files and rebuilds on any change.
```
ChrisBrownie55:
```suggestion
Acts just like `microbundle build`, but watches your source files and rebuilds on any change.
```

README.md
```diff
@@ -39,13 +44,41 @@ Microbundle includes two commands - `build` (the default) and `watch`. Neither r
...
 
+### Using with typescript
+
+Just point to a `.ts` file as input using either the cli or the `source` key in your `package.json` and you’re done.
```
ChrisBrownie55:
```suggestion
Just point the input to a `.ts` file through either the cli or the `source` key in your `package.json` and you’re done.
```

README.md
```diff
@@ -16,18 +16,23 @@
 - Produces tiny, optimized code for all inputs
 - Supports multiple entry modules _(`cli.js` + `index.js`, etc)_
 - Creates multiple output formats for each entry _(<abbr title="CommonJS (node)">CJS</abbr>, <abbr title="Universal Module Definition">UMD</abbr> & <abbr title="ECMAScript Modules">ESM</abbr>)_
+- 0 configuration Typecript support
```
ChrisBrownie55:
```suggestion
- 0 configuration TypeScript support
```

ChrisBrownie55: You could also say "No" instead of "0" but that's just choice of words.

README.md
```diff
@@ -39,13 +44,41 @@ Microbundle includes two commands - `build` (the default) and `watch`. Neither r
...
-(the root module in your program) from the `source` field in your `package.json`. It will infer the output directory and filename(s) from the `main` field. For UMD builds, microbundle will use a snake case version of the `name` field in your `package.json` for the export name; you can also specify a name via an `amdName` field or the `name` CLI option.
+Unless overridden via command line, microbundle uses the `source` property in your `package.json` to locate the input file, and the `main` property for the output.
+
+For UMD builds, microbundle will use a snake case version of the `name` field in your `package.json` as export name. This can be overridden either by providing an `amdName` key in your `package.json` or via the `--name` flag in the cli.
 
 ### `microbundle watch`
 
 Just like `microbundle build`, but watches your source files and rebuilds on any change.
 
+### Using with typescript
```
ChrisBrownie55:
```suggestion
### Using with TypeScript
```

README.md
```diff
@@ -39,13 +44,41 @@ Microbundle includes two commands - `build` (the default) and `watch`. Neither r
...
+### Specifying builds in `package.json`
+
+You can specify output builds in a `package.json` as follows:
+
+```
+"main": "dist/foo.js",          // CJS bundle
+"umd:main": "dist/foo.umd.js",  // UMD bundle
+"module": "dist/foo.mjs",       // ES Modules bundle
+"source": "src/foo.js",         // custom entry module (same as 1st arg to microbundle)
+"types": "dist/foo.d.ts",       // Typescript typings
```
ChrisBrownie55:
```suggestion
"types": "dist/foo.d.ts",       // TypeScript typings
```

ChrisBrownie55:
If we're modifying the README here could we also address the fact that [`--define` and `--alias` are missing from the README](https://github.com/developit/microbundle/pull/312#issuecomment-473600805)

Basically, it would be nice to update the `All CLI Options` section with the updated output of `microbundle --help`, as well as, provide a short example of how to use `--define` and `--alias` because it isn't currently apparent.

isthatcentered:
@ChrisBrownie55 definitely. 

The cli already provides some nice to the point examples, do you want to use those or do you want to add something a bit more descriptive ?

Here's the cli output
image

README.md
```diff
@@ -16,18 +16,23 @@
...
+
 `npm i -D microbundle`
 
-... then add the scripts to your `package.json`:
+### Set up your `package.json`
 
 ```js
 {
+  "source": "src/foo.js",         // Your source file (same as 1st arg to microbundle)
+  "main": "dist/foo.js",          // Desired output path
```
developit:
```suggestion
  "main": "dist/foo.js",        // output path for CommonJS/Node
  "module": "dist/foo.mjs",     // output path for JS Modules
  "unpkg": "dist/foo.umd.js",   // optional, for unpkg.com
```

README.md
```diff
@@ -16,18 +16,23 @@
...
   "scripts": {
-    "build": "microbundle",
+    "build": "microbundle",       // uses the "source" and "main" key in the file as input and output paths by default
```
developit:
```suggestion
    "build": "microbundle",       // uses "source" and "main" as input and output paths by default
```

developit:
Great work! I left a couple little tweaks but otherwise this looks good to merge!

isthatcentered:
I feel like some informations about the "target" option could be worth adding. 

My guess from looking at the output is that "target web" exposes the module as a global variable which "target node" doesn't. 

Is that right, I did not found an equivalent on Rollup's docs but I might not be looking for the right keyword.

PR #361, opened 2019-03-22
Rollup 1.0 + other updates

ForsakenHarmony:


src/index.js
```diff
@@ -523,7 +523,7 @@ function createConfig(options, entry, format, writeMeta) {
 					useTypescript &&
 						typescript({
 							typescript: require('typescript'),
-							cacheRoot: `./.rts2_cache_${format}`,
+							cacheRoot: `./node_modules/.rts2_cache_${format}`,
```
developit: clever.

maraisr: Rekon we could take that 1 step further with,

```suggestion
							cacheRoot: `./node_modules/.rts2_cache_${format}`,
							cacheRoot: `./node_modules/.cache/.rts2_cache_${format}`,
```

package.json
```diff
@@ -42,61 +42,61 @@
...
-    "rollup-plugin-terser": "^3.0.0",
-    "rollup-plugin-typescript2": "^0.19.0",
+    "rollup-plugin-postcss": "^2.0.3",
+    "rollup-plugin-terser": "^4.0.4",
+    "rollup-plugin-typescript2": "^0.20.1",
     "sade": "^1.4.0",
     "tiny-glob": "^0.2.6",
     "tslib": "^1.9.0",
-    "typescript": ">=2.8.3"
+    "typescript": "^3.3.4000"
```
maraisr: Now that [TypeScript 3.4](https://github.com/Microsoft/TypeScript/releases/tag/v3.4.1) is a thing, do you think we could make that read;

```suggestion
    "typescript": "^3.4.0"
```

src/index.js
```diff
@@ -1,6 +1,6 @@
 import fs from 'fs';
 import { resolve, relative, dirname, basename, extname } from 'path';
-import chalk from 'chalk';
+import { green, red, yellow, white, blue } from 'kleur';
```
maraisr: :100:

src/index.js
ForsakenHarmony: was kinda annoyed with those 3 directories in my project dir 🙃

src/index.js
```diff
@@ -219,17 +225,19 @@ export default async function microbundle(inputOptions) {
...
+		steps.map(config => async () => {
+			const { inputOptions, outputOptions } = config;
 			inputOptions.cache = cache;
 			let bundle = await rollup(inputOptions);
 			cache = bundle;
-			const { code } = await bundle.write(outputOptions);
-			return await getSizeInfo(code, outputOptions.file);
+			await bundle.write(outputOptions);
+			return await config._sizeInfo;
+			// return await getSizeInfo(code, outputOptions.file, options.raw);
```
maraisr: You've commented here, why if you don't mind me asking?

ForsakenHarmony: oh that should be removed

src/index.js
```diff
@@ -16,7 +16,6 @@ import postcss from 'rollup-plugin-postcss';
 import gzipSize from 'gzip-size';
 import brotliSize from 'brotli-size';
 import prettyBytes from 'pretty-bytes';
-import shebangPlugin from 'rollup-plugin-preserve-shebang';
```
maraisr: Oh good - I seriously spent about 30 mins trying to PR that project with updates. And I could not get rollup to like anything I was doing in that `transformChunk` hook. 👀

ForsakenHarmony:
Oh yes, seems like I forgot to push that commit

PR #365, opened 2019-03-29
add example to the readme

horacioh:
I created a basic example using microbundle with:
- React
- Typescript
- Storybook
- Styled systems
- styled components

I thought maybe is useful to someone else?


ChrisBrownie55:
I don't think it makes sense to keep your example listed under the *"Built with Microbundle"* section as that seems to be mostly for packages built with Microbundle. You might add an examples section in the readme.

However, it might be better to create examples inside this repository for a couple reasons:

1. Easier to PR changes and keep track of issues that need to be changed all in one place.
2. Links can end up broken so keeping the code in an examples folder would be a better idea.

horacioh:
I think you are right. I can create the examples folder and add this as is :)

horacioh:
closing this in favor of #365

PR #366, opened 2019-03-29
Add example folder

horacioh:
add an example folder to share project configurations using microbundle.

the example I added is using:
- React
- Typescript
- Storybook
- Styled systems
- styled components

hopefully this will be helpful for new devs :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,7 @@ Here's what's coming up for Microbundle:
 - [react-hooks-lib](https://github.com/beizhedenglong/react-hooks-lib) A set of reusable react hooks.
 - [mdx-deck-live-code](https://github.com/JReinhold/mdx-deck-live-code) A library for [mdx-deck](https://github.com/jxnblk/mdx-deck) to do live React and JS coding directly in slides.
 - [react-router-ext](https://github.com/ri7nz/react-router-ext) An Extended [react-router-dom](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-dom) with simple usage.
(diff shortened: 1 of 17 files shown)


maraisr:
Thanks for the PR, but what problem are you trying to solve here? Storybook is still using webpack to build it's things. So really all you're showing here is that `microbundle` _can_ be used alongside Storybook. What would be cool, would be if Storybook was to consume the microbundle dev build. But even then this is a bundler...

Maybe this can be an example you could maintain in your own personal land, and we can ask @developit if we could reference it in the readme or something.

All in all, the maintainers here wouldn't want to now maintain examples, and test that all the examples still work etc...

horacioh:
thanks for the response @maraisr !

my first approach was what you said: [https://github.com/developit/microbundle/pull/365](https://github.com/developit/microbundle/pull/365). I think this is cleaner for the proyect too.

I'm happy to reopen/update/delete this PRs. I found examples really useful for newcomers to the project, and check how the setup (specially the setup) is done is really helpful.

Hope we can give the community a way to publish more examples as this one!

Let me know what you think 👍

developit:
I'd love a microbundle-examples repo.

devinrhode2:
> Storybook is still using webpack to build it's things. So really all you're showing here is that microbundle can be used alongside Storybook. What would be cool, would be if Storybook was to consume the microbundle dev build.

Storybook now has a vite builder, which _can_ replace webpack. Vite is mostly based on Rollup, with some esbuild sprinkled in where possible.

I wonder if Microbundle could be implemented on top of vite?

PR #370, opened 2019-04-01
Adds a `--css-modules` option

maraisr:
With this PR, I aim to add an option to the cli, that'll allow developers using css-modules to bundle their things, using this. I gather that this would be solved in the future with a config file, or convention.

I aim for something like `microbundle entry.js --css-modules` that when ran to produce this, as yeah, probably not many people want this as the default. _And_ also makes this backwards compatible.


maraisr:
Starting this draft here - the snapshots are failing for me, even at a _clean_ clone. So will investigate. Also wanting to confirm a few things, and maybe clean that config setting thing up a bit. Also, the jest fixtures are a bit awkward for this. So going to also split out cli args into potentially separate tests, that run "build commands" from the package json in its own fixture or something. #minddump :100:

maraisr:
The unit tests for `esnext-ts` are failing, for me locally as well. I can't even get _master_ to snapshot without failing on that test.

It seems there is no lock file for this project, do we think that maybe one of the dependants (TypeScript 3.4) could have caused issues? Master hasnt built and ran in over 10 days?

src/index.js
```diff
@@ -470,6 +470,9 @@ function createConfig(options, entry, format, writeMeta) {
 									preset: 'default',
 								}),
 						].filter(Boolean),
+						modules:
+							!!options['css-modules'] &&
+							(!!options.watch || { generateScopedName: '_[hash:base64:5]' }),
```
developit: Think we could use `--css-modules` here as the scoped name if its a String? (`--css-modules "_[hash:base64:5]"`)?

```suggestion
							(!!options.watch || {
								generateScopedName: typeof options['css-modules']=='string' ? options['css-modules'] : '_[hash:base64:5]'
							}),
```

src/index.js
```diff
@@ -470,6 +470,9 @@ function createConfig(options, entry, format, writeMeta) {
 									preset: 'default',
 								}),
 						].filter(Boolean),
+						modules:
```
developit: We should also document that `microbundle` already supports `foo.module.css` for enabling CSS modules per-file, since [rollup-plugin-postcss' autoModules option](https://github.com/egoist/rollup-plugin-postcss#automodules) defaults to `true`.

In that case we could potentially turn `--css-modules` into a tristate flag (where the default value is `null`):

| Command | What happens when it is run |
|-|-|
| `microbundle` | Allow CSS modules via `.module.css`
| `microbundle --css-modules` | Turn on CSS modules for all `.css` files
| `microbundle --css-modules false` | Disable CSS modules, even for `.module.css` files
| `microbundle --css-modules "_[hash]"` | Turn on CSS modules for all `.css` files<br>**and** provide custom name generation

maraisr:
@developit thanks for that - the PR has been updated! 🚀

transitive-bullshit:
It'd be great for this to be added to `microbundle`, but either way the ability to use `*.module.css` to enable CSS modules with the current version isn't easy to find or figure out.

I'd love to make sure the docs get updated accordingly with a note about CSS module support. Happy to send a PR once this gets merged.

maraisr:
_if_ this ever gets merged ;) @transitive-bullshit

src/prog.js
```diff
@@ -33,6 +33,11 @@ export default handler => {
 		.option('--name', 'Specify name exposed in UMD builds')
 		.option('--cwd', 'Use an alternative working directory', '.')
 		.option('--sourcemap', 'Generate source map', true)
+		.option(
+			'--css-modules',
+			'Turns on css-modules for all .css imports. Passing a string will override the generatedScopeName. eg --css-modules="_[hash]"',
+			null,
```
ForsakenHarmony: setting the default to true would make the conditional above more readable 🙃

developit: @ForsakenHarmony it's a tri-state though. `--css-modules true` turns on CSS modules for all `.css` files, whereas the default `null` value turns on CSS modules only for `.module.css` files.

I wonder though if you're pointing out that there might not be much reason to offer `--css-modules false`? (which disables css modules for `.module.css` files)

test/fixtures/basic-css-modules/src/index.js
```diff
@@ -0,0 +1,7 @@
+import twoStyles from './two.css';
```
hbroer: this is called `twoStyles` but contains only one definition. Was it intended to test if tree shaking works?

developit: there's no tree shaking for CSS, not sure why it was called `two`.

hbroer: ah ok, i thought it could be made. Checked it, even webpack does not remove unused css modules code (but they could because they know all referencing code) ^^

LeoVS09:
Guys, we want to move our projects on microbundle, but cannot do it without css modules. Are you stop developing this feature?

maraisr:
What is left to get this across the line?? @developit @ForsakenHarmony @transitive-bullshit 

Its just docs yeah?

src/index.js
```diff
@@ -513,6 +513,25 @@ function createConfig(options, entry, format, writeMeta) {
...
+							options['css-modules'] == 'false' ||
+							options['css-modules'] === null
+								? false
+								: {
+										generateScopedName:
+											options['css-modules'] === 'true' ||
+											options['css-modules'] === true
+												? '_[hash:base64:5]'
+												: options['css-modules'],
+								  },
```
ForsakenHarmony: let's pull this out, it's not very readable

ForsakenHarmony:
Sorry about the delays, mental health not doing so well

But yeah would be great adding something to the readme

maraisr:
Hi @ForsakenHarmony 

I have updated this PR, with a focus on readability. Wrote some docs on how to consume this, and also a better explanation on how to consume this feature.

One change however is, to treat `.module.css` as modules by defaults, rather than the _current_ implementation where _nothing_ is a module.

@hbroer i wrote tests for all avenues.

I will however note, that current the `--css-modules null` test, is incorrect. Because there is no way pass scope names to the underlying plugin, without also making it behave as though it was a `--css-modules true`. The comment around what the scope names will be _will_ be accurate, and am raising a PR to the plugin to fix this. See below as to why. 

https://github.com/egoist/rollup-plugin-postcss/blob/3fc6c44e7fbd287b1a295d5e3770e3e6cbaa7731/src/postcss-loader.js#L75-L97

A simple solution for now, would be to not pass the scopeName config when the option is null.

test/fixtures/css-modules--false/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "css-modules--false",
+  "scripts": {
+    "build": "microbundle --no-sourcemap --css-modules false"
+  }
+}
```
ForsakenHarmony: Minor nit, a lot of these files don't seem to have a final newline, does your editor not respect the editorconfig?

maraisr: Sorry my bad, I assumed prettier took care of that. Will update with new lines.

ForsakenHarmony: Oh, do we have conflicting editorconfig and prettier configs?

maraisr: No sorry! Saw prettier and assumed it was used... But actually, its just running over the markdown files.

https://github.com/developit/microbundle/blob/master/package.json#L30-L31

Will fix these newline issues none the less.

maraisr: Here's a suggestion, why don't you approve this PR. Get it merged, and I'll raise a separate PR that runs prettier over all source files + add's in the husky pre-commit hook for that.

ForsakenHarmony: Sure

VincentCharpentier:
Thanks for this feature 👍 

Is it possible to have it in a release soon ?

PR #386, opened 2019-05-03
refactor(jest): use it.each & remove sourcemaps from snapshots

wardpeet:
While working on the bublé => babel conversion the source-maps snapshot was actually cluttering up the diffs and I'm unsure why they are there. I changed the tests to omit the source-maps from snapshots but  I do check file count to make sure they are created.


test/index.test.js
```diff
@@ -55,64 +55,70 @@ const parseScript = (() => {
 })();
 
 describe('fixtures', () => {
+	const dirs = [];
 	fs.readdirSync(FIXTURES_DIR).forEach(fixtureDir => {
```
ForsakenHarmony: this could probably be a map + filter for a more functional approach

test/index.test.js
```diff
@@ -55,64 +55,70 @@ const parseScript = (() => {
...
-						'Directory tree:',
-						printedDir,
-						strip(output),
-					].join('\n\n'),
-				).toMatchSnapshot();
+		dirs.push(fixtureDir);
+	});
 
-				fs.readdirSync(resolve(dist)).forEach(file => {
+	it.each(dirs)(
```
ForsakenHarmony: didn't know this was a thing 👌

test/index.test.js
wardpeet: 👍  make sense!

ForsakenHarmony:
nice

PR #395, opened 2019-05-25
fix: allow disabling minification from cli

wardpeet:
It seems like adding `--compress=false` doesn't work because we never normalize "true"|"false" to a boolean. Now I normalize "false or "0" to a boolean so the compress can be set from the cli.

I don't think many people need this feature but it's easy when debugging output.


ForsakenHarmony:
hmm ye the patch update of async promises broke our tests

PR #362, opened 2019-03-24
chore: convert bublé to babel for es2015 compilation

wardpeet:
Remove bublé and use babel instead. This gives us the opportunity to easily move to modern builds. The file sizes are all kept almost the same. The sizes are raw bytes which means gzip and brotli isn't used so the diff is going to be minimal.

I also added 2 scripts so we can generate these tables easier

Latest size report from travis:
https://travis-ci.org/developit/microbundle/builds/537119810#L330-L380

<details>
<summary>File sizes</summary>

| ﻿file                                           | bublé   | babel    |
|------------------------------------------------|---------|----------|
| alias/alias-mapping.js                         | 42      | 42       |
| alias/alias-mapping.mjs                        | 42      | 42       |
| alias/alias-mapping.umd.js                     | 121     | 121      |
| async-iife-ts/async-iife-ts.js                 | 116     | 116      |
| async-iife-ts/async-iife-ts.mjs                | 117     | 117      |
| async-iife-ts/async-iife-ts.umd.js             | 199     | 199      |
| async-iife-ts/index.d.ts                       | 0       | 0        |
| async-ts/async-ts.js                           | **104** | 120      |
| async-ts/async-ts.mjs                          | **108** | 132      |
| async-ts/async-ts.umd.js                       | **287** | 303      |
| async-ts/index.d.ts                            | 62      | 62       |
| basic/basic-lib.js                             | **468** | 502      |
| basic/basic-lib.mjs                            | **468** | 502      |
| basic/basic-lib.umd.js                         | **643** | 677      |
| basic-css/basic-css.css                        | 48      | 48       |
| basic-css/basic-css.js                         | 133     | 133      |
| basic-css/basic-css.mjs                        | 133     | 133      |
| basic-css/basic-css.umd.js                     | 308     | 308      |
| basic-json/basic-json.js                       | 168     | **113**  |
| basic-json/basic-json.mjs                      | 168     | **113**  |
| basic-json/basic-json.umd.js                   | 344     | **289**  |
| basic-multi-source/a.js                        | 53      | 53       |
| basic-multi-source/a.mjs                       | 54      | 54       |
| basic-multi-source/a.umd.js                    | 136     | 136      |
| basic-multi-source/b.js                        | 53      | 53       |
| basic-multi-source/b.mjs                       | 54      | 54       |
| basic-multi-source/b.umd.js                    | 136     | 136      |
| basic-ts/basic-lib-ts.js                       | **130** | 147      |
[...]

PR #347, opened 2019-03-11
Preserve #__PURE__ annotations

Andarist:
fixes #346

PR #396, opened 2019-05-25
 feat(babel): merges babelrc with microbunde babel config

wardpeet:
Enables custom babelrc files to get merged with our babelrc file. By default babel already does this if presets/plugins do not exists. With this PR we merge them together.

For `@babel/preset-env` I've done a special merge as we want to make sure `transform-async-to-generator` is excluded and modules is set to false.

This allows people to set `@babel/preset-env` with babel-polyfill as many applications do. This is less relevant for libraries unless they want to ship a few polyfills magically.

Another change that has been added is we only run the node_modules babel transform when the `defines` option is set. (I think it should speed up microbundle a bit if people aren't using it)


test/__snapshots__/index.test.js.snap
```diff
@@ -658,11 +705,105 @@ exports[`fixtures build class-properties with microbundle 5`] = `
...
+5.05 kB: custom-babelrc.js.br
+5.6 kB: custom-babelrc.mjs.gz
+5.06 kB: custom-babelrc.mjs.br
+5.67 kB: custom-babelrc.umd.js.gz
+5.12 kB: custom-babelrc.umd.js.br"
+`;
+
+exports[`fixtures build custom-source with microbundle 2`] = `6`;
+
+exports[`fixtures build custom-source with microbundle 3`] = `
```
wardpeet: there is a bug in core-js@3 where es.promise get transpiled with `{esmodules:true}` which I'll try to fix upstream. When using core-js@2 it's much cleaner. It's not a microbundle issue

ForsakenHarmony:
So wait, if someone puts a babelrc in their project dir it gets overridden with our defaults?

wardpeet:
Let me try to explain a bit better.
So by default we will add `'@babel/plugin-transform-react-jsx', 'babel-plugin-transform-replace-expressions', 'babel-plugin-transform-async-to-promises', '@babel/plugin-proposal-class-properties'` plugins even if a custom babelrc file is added. The options of each plugin can be overridden if the same plugin is used inside the custom babelrc file.
https://github.com/developit/microbundle/blob/555af26edba16ca3066c66b94e93a94c62743ae1/src/lib/babel-custom.js#L55-L76

The same principle goes for `@babel/preset-env` with the only caveat that we don't allow people to overwrite the `modules` option and we always exclude `transform-async-to-generator`.
https://github.com/developit/microbundle/blob/555af26edba16ca3066c66b94e93a94c62743ae1/src/lib/babel-custom.js#L91-L112

Does this explains it better?

ForsakenHarmony:
yeah that explains it, ideal would be if we could have a preset like

https://github.com/developit/microbundle/pull/362#discussion_r285870535

but I guess we'll want to override at least the modules stuff either way

wardpeet:
I don't think we need a babel preset. I don't see a big issue with merging what we do now. The only problem that I might see is we always exclude `transform-async-to-generator` which is something I don't mind dropping.

ForsakenHarmony:
not sure if that's really a problem either

ForsakenHarmony:
hmm, broke tests with #347

wardpeet:
let me fix this :)

ForsakenHarmony:
it's a bit weird, not sure if I should just update the snapshots

alexandernanberg:
The issue I can see with merging the config is that it's a bit "magical" and ambiguous. With a custom babel preset it would be very clear what's happening and the end user could still customize the other preset options (exclude etc). There are no cases where you want to be able to completely override the default config?

Anyway great work on this! 😄

briancodes:
I'm using `microbundle` and loving the zero config. But.... in order to get Jest working it seems I need to supply a `.babelrc` with the following: 

```
{
  presets: ['@babel/preset-env']
}
```
Would the merging of this `bablerc` (which I only need for tests) cause any adverse effects do you think e.g. I'd rather just use microbundle defaults, but don't see a way around using a config file in this instance

wardpeet:
@briancodes you mostly want to use babel env to fix this. There are 2 options setting BABEL_ENV=test or NODE_ENV=test.

in babelrc you can create this config:
```js
{
  "env": {
    "test": {
      presets: ['@babel/preset-env', { "node": "current", "modules": false }]
    }
  }
}
```

PR #431, opened 2019-06-24
Fix duplicate package name warning

andrewiggins:
Some tests had duplicate package names. I updated those packages and relevant snapshots to have unique names


ForsakenHarmony:
thx

PR #444, opened 2019-07-08
fix: add Flow support back

### LINKED ISSUE #443: Missing Flow support on `0.12.0-next.*` versions?
I see the Flow babel plugins are in `package.json` but they are not referenced anywhere in the code.

If I try to build my code with `0.12.0-next.3` it throws a syntax error on the first `import { type Smth } from '...'`.

FezVrasta:
fixes #443


src/lib/babel-custom.js
```diff
@@ -64,6 +64,9 @@ export default babelPlugin.custom(babelCore => {
 						pragma: customOptions.jsx || 'h',
 						pragmaFrag: customOptions.jsxFragment || 'Fragment',
 					},
+					!customOptions.typescript && {
```
FezVrasta: I copied this from the previous Microbundle version, there we run the Flow plugin anytime it's not TypeScript.

ForsakenHarmony:
I think it was initially skipped because typescript kinda won, but I'm fine with merging this

PR #448, opened 2019-07-16
Switch from .mjs to .esm.js to avoid webpack tomfoolery

developit:
☹️


ForsakenHarmony:
I thought we switched back ages ago

ForsakenHarmony:
gotta update the snapshots 👀

PR #458, opened 2019-07-30
feat: add support for custom tsconfig path

chuahcheeshian:
With `tsconfig.json` [extends](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends) option, it is usually normal for a project to have multiple `tsconfig.json`s for different purposes, such as test, build, IDE IntelliSense, etc.

With this PR, a `tsconfig` option from cli can be passed down to `rollup-plugin-typescript2` when defined.


ForsakenHarmony:
Thanks for the PR 👌

Issue #477, opened 2019-09-12
Officially support more elaborate externals


Xiphe:
_This is a feature request/Proposal_

As a user of microbundle I want ways to configure which imports are to be considered external.
In order to reduce my bundle size and also to work with non-standard imports (images, fonts, ...) that are not supported by microbundle.

## Status Quo

1. As described in https://github.com/developit/microbundle/issues/283#issuecomment-523879163, there is a (I guess unoffical) way to use regex in the `--exclude` flag, which can be used to have rollup treat specific import statements as external.
2. Presence of the `--exclude` flag triggers if pkg.dependencies are treated as external or not (see [index.js#L387](https://github.com/developit/microbundle/blob/master/src/index.js#L387-L388)) I find that quite confusing and the behaviour is not documented. 

## Problem

Given I want to use `--external .*/assets/.*` to work with non-standard imports
And I also want all my dependencies to be excluded
Then I need to add all dependencies to `--external` causing code duplication and manual overhead.

## Proposal

In order to fix the problem and reduce confusion I propose to have it work like this:

### `$ microbundle`

Exclude `dependencies`, `peerDependencies` 

### `$ microbundle --external .*/assets/.*`

Exclude `dependencies`, `peerDependencies` and `.*/assets/.*`
And explicitly support regular expressions or globs under `--external`

### `$ microbundle --external .*/assets/.*,angular --include-dependencies`

Exclude `peerDependencies` and `.*/assets/.*`
All normal dependencies ecxept `angular` are included in the bundle

## Non-breaking alternative

I realize that the above proposal is really tied to my personal usecase of (miss-?)using `--external` for non-standard imports.
It would also introduce a potential pretty impactful breaking change (since users of microbundle might ship bundles that lack bundled dependencies).

Alternatively the current behaviour could stay as is, using regex in `--external` will be considered hacky. And we introduce a new, documented way to exclude assets:

### `$ microbundle --external-glob **/*.(jpg|png)`

Exclude `dependencies`, `peerDependencies` and all imports matching `**/*.(jpg|png)`

### `$ microbundle --external angular --external-glob **/*.(jpg|png)`

Exclude `peerDependencies`, `angular`, and all imports matching `**/*.(jpg|png)`

---

Let me know what you think about this. I would be really happy to implement the changes and provide a PR.

Xiphe:
Yeah! i like the approach!

I find it confusing to wrap a glob with `/` since it then looks more like regex to me, I would therefore propose the following:

[...]

PR #478, opened 2019-09-16
support regex in --external

Xiphe:
in order to support ignoring of file types not supported by microbundle

PTAL: @developit 

fix https://github.com/developit/microbundle/issues/477


src/index.js
```diff
@@ -382,19 +382,31 @@ function createConfig(options, entry, format, writeMeta) {
...
 		// bundle everything (external=[])
 	} else if (options.external) {
-		external = external.concat(peerDeps).concat(options.external.split(','));
-	} else {
-		external = external
+		let containsPackage = false;
+		external = options.external
+			.split(',')
+			.map(ext => {
+				const match = ext.match(/^\/(.*)\/(i?)$/);
```
developit:
```suggestion
				const match = ext.match(/^\/(.+)\/(i?)$/);
```

Xiphe:
Going to update the branch and try to fix the tests

Xiphe:
Rebased on master(https://github.com/developit/microbundle/commit/951a23560d8ac2fc0d846a7505ac03339552abc7). The failing tests are the same as on master. Looks like some snapshots are outdated. But the changes seem ok.

Xiphe:
Rebased on master (c03b535cd61f21cb25275ae315252fea69d1cb96). Is there anything I can do to help getting this into master?

Xiphe:
I still have issues running the tests locally, I have 13 failing snapshots and no Idea what I did wrong. (ref https://github.com/developit/microbundle/issues/542) 

I would really like to help bringing this project forward but struggle to get everything to work as expected. (currently on node@12 / OSX@10.14)

Xiphe:
Rebased on master (c03b535cd61f21cb25275ae315252fea69d1cb96). Is there anything I can do to help getting this into master?

src/prog.js
```diff
@@ -25,7 +25,11 @@ export default handler => {
 		.option('--format, -f', 'Only build specified formats', DEFAULT_FORMATS)
 		.option('--watch, -w', 'Rebuilds on any change', false)
 		.option('--target', 'Specify your target environment (node or web)', 'web')
-		.option('--external', `Specify external dependencies, or 'none'`)
+		.option(
+			'--external',
+			`comma-separated list of regex and module-names, or 'none'`,
```
developit: [nit] can we reverse these so that "module names" comes first (because it's simpler/more common)?

```suggestion
			`List of module names or regular expressions to leave as external dependencies, or 'none' to inline everything.`,
```

src/prog.js
```diff
@@ -25,7 +25,11 @@ export default handler => {
...
+		)
+		.example('microbundle --external /\\.jp(e)?g$/i,some_package')
```
developit:
```suggestion
		.example('microbundle --external some_package,/\\.jp(e)?g$/i')
```

developit:
Just a question that arose as I tried to tweak the docstrings for this:  maybe Regular Expressions on the CLI is going a little too far down the configuration road?

We could probably get away with allowing glob/minimatch patterns here instead, since that would automatically support the direct name matching:

```sh
microbundle --external some_module,*.jpe?g
```

Xiphe:
@developit Thanks for the feedback. Will update the PR later today. 

As with the glob vs regex. In our earlier discussion you proposed to have globs with wrapping slashes (https://github.com/developit/microbundle/issues/477#issuecomment-531081702) in order to not have to introduce an additional `--external-glob` flag while still not accidentally globbing for whole module names.

I said `some_module,/*.jpe?g/` looks a lot like regex so why not directly use regex.

I personally don't care if we use regex or globs and will happily tweak this PR in any direction you want. I think the one important thing is that I as a user need to be sure whether my `--external` is interpreted as a module name or as a matcher.

Xiphe:
While thinking about how to make globs possible i got another idea:

We could borrow the [protocol syntax from yarn](https://yarnpkg.com/features/protocols) and then declare globs like this: `my-module,glob:*.jpe?g`. That would leave a lot of space open for future innovation inside the external flag (`regex:/\.custom$/i`, etc).

In addition we could also allow `\,`-escaping the colons in order to support brace expansions `--external="glob:foo.{ts\,js}"

Let me know what you think. For now I'll update this without a feature change.

Xiphe:
PTAL @developit – Added your suggestions.

rmarskell:
Any idea when this PR might get reviewed/merged @ForsakenHarmony ? Would love to see this added.

developit:
Hiya - sorry for dropping the ball on this. I've tried to resolve merge conflicts since apparently we merged a different regex externals implementation at some point. The only thing I wasn't able to keep is support for case-insensitive matches - we merge the regexes together into a single regex, and doing case-insensitive subset matching is a huge pain. Not sure what to do there.

Xiphe:
No worries and thanks for the update. From what I remember where I would have needed this case sensitivity was a nice to have but not required.

Issue #487, opened 2019-10-01
Requires node >= 10.16


transitive-bullshit:
`microbundle` fails to install on node 8 and some versions of node 10 due to its relatively minor dependency on [brotli-size](https://github.com/erwinmombay/brotli-size). (see this [PR](https://github.com/erwinmombay/brotli-size/pull/11) where the requirement was added)

I tracked this down due to an [error in CI](https://travis-ci.com/saasify-sh/saasify/jobs/240609522) where we need to support `node >= 8`.

Given that Node.js `v8` and minor versions of Node.js `v10` before `v10.16` are still in LTS, I'd propose making `brotli-size` an optional dependency until they're deprecated or at least less common.

At the very least, I'd suggest making this requirement explicit in the readme and/or package.json `engines` field.

Mucho gracias for this awesome package btw! 😄

lizy0329:
嗯，here is nodejs version is <= 10.16 Error

![WeChat Work Screenshot_20191106095506.png](https://i.loli.net/2019/11/06/b7WB5XvCzSjEfRs.png)

ForsakenHarmony:
Released in `0.12.0-next.7` sorry it took so long

transitive-bullshit:
Thanks for all your OSS work @ForsakenHarmony 🙏

PR #492, opened 2019-10-14
wrap brotli in try/catch to avoid error for lower versions of node

madalynrose:
Addresses #487 by catching any error for `brotliSize` and not displaying stats for brotli.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -119,12 +119,19 @@ async function getSizeInfo(code, filename, raw) {
 		'gz',
 		raw || code.length < 5000,
 	);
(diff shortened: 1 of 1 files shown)


developit:
Since master tests are failing I'm going to merge this.

developit:
Thanks so much @madalynrose! We'll get an update published ASAP so Gatsby can switch back to the main Microbundle package.

ForsakenHarmony:
Really sorry it took so long, it's in 0.12.0-next.7

0.12 should be out soon as well

PR #494, opened 2019-10-17
Add github actions

ForsakenHarmony:


.github/workflows/nodejs.yml
```diff
@@ -1,20 +1,19 @@
...
+      - staging.tmp
 
 jobs:
   build:
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x]
+        node-version: [8.x, 10.x, 12.x, 13.x]
```
marvinhagemeister: Node 8.x is EOL'ed starting with this year. We can remove it🎉

kristoferbaxter: Yep, and all code specifically targeting it.

ForsakenHarmony:
I guess sizereport doesn't give us much here

PR #499, opened 2019-10-19
Added editorconfig properties for preserving trailing newline

### LINKED ISSUE #497: Preserve trailing newline in mangle.json
See https://github.com/preactjs/preact/blob/e22cc99ca398bb0a417ccc151a520c41215afe55/.editorconfig#L17-L22

brynmrk:
Fixes #497

### DIFF
--- a/test/__snapshots__/index.test.js.snap
+++ b/test/__snapshots__/index.test.js.snap
@@ -22,8 +22,8 @@ Build \\"aliasMapping\\" to dist:
 46 B: alias-mapping.js.br
 62 B: alias-mapping.esm.js.gz
 46 B: alias-mapping.esm.js.br
-118 B: alias-mapping.umd.js.gz
-83 B: alias-mapping.umd.js.br"
+120 B: alias-mapping.umd.js.gz
+85 B: alias-mapping.umd.js.br"
 `;
 
 exports[`fixtures build alias with microbundle 2`] = `3`;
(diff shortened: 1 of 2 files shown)


developit:
Hi @brynmrk! Thanks for the PR! This is what we'll want to add to repositories that are built with Microbundle, but I think in order to fix #497 there may be one additional thing required: we need to check for the extra trailing line ending when reading in the `mangle.json` file (on [this line](https://github.com/developit/microbundle/blob/b170137f8df4eae74beeab22289ea576cdbb144a/src/index.js#L465)), and then insert it when writing the file (on [this line](https://github.com/developit/microbundle/blob/b170137f8df4eae74beeab22289ea576cdbb144a/src/index.js#L606-L609)).

Think you'd be interested in add that to your PR? Cheers!

test/__snapshots__/index.test.js.snap
```diff
@@ -22,8 +22,8 @@ Build \\"aliasMapping\\" to dist:
 46 B: alias-mapping.js.br
 62 B: alias-mapping.esm.js.gz
 46 B: alias-mapping.esm.js.br
-118 B: alias-mapping.umd.js.gz
```
developit: I've updated the tests now that #495 is merged, so only two should be failing. We can actually skip updating the snapshots for that in this PR since it's a real bug still to be fixed elsewhere :)

Issue #526, opened 2019-11-20
'--visualize' flag to run rollup-plugin-visualizer?


AlexHayton:
I would like to propose a PR for creating a new `--visualize` CLI flag, which will add the rollup-plugin-visualizer to the toolchain. But wanted to get some feedback before doing it.

This'll really help me, I want to confirm I got all my peerDependencies and externals correct. I wanted to get some feedback on the idea

* Is it worth doing?
* Any qualms about the naming of the CLI flag?
* Anyone already started doing it?

marvinhagemeister:
That's a cool idea, I'd love to have something like that to inspect bundles 👍 Afaik there is nobody working on that, so your good to go! The cli flag name is fine, we can bikeshed later if someone is unhappy with that (doubt it) 🙂

AlexHayton:
@marvinhagemeister I have created a PR, would you be able to give some feedback on it?

One thing I observed, a lot of snapshots changed which I wasn't expecting...

AlexHayton:
Hmm, I really can't get my snapshots to match. Does anyone have any idea?

PR #528, opened 2019-11-23
Added a --visualize option to generate build output stats

AlexHayton:
Following up from https://github.com/developit/microbundle/issues/526

Updated readme
Added the new visualize flag
Added a test with a postbuild script that copies the stats.html
Allow tests to have a postbuild script! :)


AlexHayton:
Hmm, I really can't get my snapshots to match travis. Running on OSX here. Does anyone have any idea what's going on?

test/fixtures/visualize/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "visualize",
+  "scripts": {
+    "build": "microbundle --visualize",
+    "postbuild": "mv stats.html dist/"
```
marvinhagemeister: I wonder if we can remove the need for a `postbuild` task. The docs of `rollup-plugin-visualize` don't mention it, but you can pass a directory with the `filename` setting.

https://github.com/btd/rollup-plugin-visualizer/blob/41aeee4741ab424c8e4e1ab48930ca5cbd2d8fb1/plugin/index.js#L129

AlexHayton: Sounds like a good plan.

@marvinhagemeister Do you have any idea about the other (unrelated) snapshots that are failing?
Last time I checked master is also failing too... so not sure what to do next.

developit:
@AlexHayton sorry for the slow reply - I also have the issue with snapshots being out of sync on Mac OS. Unfortunately, I haven't yet found a fix for this :/

AlexHayton:
Sorry I've been away from this for a while. Life got in the way. I'll try and resubmit soon. I worked out why the snapshots were failing, it is an issue with the randomly generated CSS class names in the visualiser output. So I am hoping it's possible to pass a specific seed or just take those out of the snapshots...

AlexHayton:
> @AlexHayton sorry for the slow reply - I also have the issue with snapshots being out of sync on Mac OS. Unfortunately, I haven't yet found a fix for this :/

@developit I think found a way around this, I will submit it seperately...

developit:
Just an update: this should all work without the Docker shennanigans now that we're on GH actions. Sorry!

AlexHayton:
Wow, zombie thread back from the dead 😄 
I'll give it another go, thank you!

AlexHayton:
Actually reading my past self's comments, the specific blocker here was the CSS class names...

developit:
No pressure btw! I like the idea of this feature though, and I'd be happy to rebase - I don't have the snapshots issues when working in the repo anymore (perhaps they're still there...)

rschristian:
Ended up implementing in #549 rather than editing this PR directly. It's finally merged!

PR #532, opened 2019-12-07
Defensive code for accessing nested object, options

realdennis:
Error

To fix security nested dependencies, I just try the `v0.12.0-next6` to build my library, and I found this issue.

`preset.options` might be `undefined`, I add a defensive code.


ForsakenHarmony:
Thanks for the contributions 🙌

PR #548, opened 2020-01-22
Strip comments from output

developit:


developit:
@marvinhagemeister Ideally I would like to extract license comments into a file if they need to be preserved.

jnordberg:
I'd love to see this merged, half my bundle is a microsoft license comment from something

marvinhagemeister:
@developit I have no permissions to check the GH actions result. Not sure why it's stuck.

developit:
Waiting on PR approval :P

PR #549, opened 2020-01-24
GitHub actions

ForsakenHarmony:
Had to recreate because after merging the other pr first the branch was deleted and github is too stupid to allow changing the target now


developit:
This is awesome! I wonder if we should switch from travis-size-report to one that works with Actions?

ForsakenHarmony:
I suppose it could be nice to show those in some way, but not necessary

Issue #554, opened 2020-01-29
Provide a Jest Transform


therealparmesh:
It may be convenient to expose something to consumers that will make it easier to use Jest - something like the default babel options (presets, plugins, etc.) or a transform module (see something like https://github.com/ActuallyACat/jest-esm-transformer).

This makes using syntax like `import`/`export` much easier for folks who use this amazing library for bundling and Jest for tests.

Thank you!

developit:
@therealparmesh great suggestion. I see that Jest can already do this when Babel is installed, though because Microbundle abstract Babel Jest likely doesn't detect it. So I guess you're saying it would be nice if there was a Microbundle-provided option here?

If so, I agree!

therealparmesh:
Precisely what I meant!

FWIW, my current workaround is to have my Jest tests import from my built JS (You can see it here: https://github.com/therealparmesh/object-to-formdata/blob/b0c0be74a3e510ea8be2efa56f85e4e21ba3afa7/src/index.test.js#L1). This works well - even with features like code coverage (because of the source maps that point to the source file).

therealparmesh:
Any feedback on this? 😄

hbroer:
microbundle could save the configuration to the project root with a "This is autogenerated. Don't change it."

If microbundle would come with a "public" Babel config, it would also be possible to put a eslint config to the root with just a `extend` property.

xavierfoucrier:
Hi @developit,

I am using `jest` with `microbundle` to test a package source code. I was able to use `import/export` syntax until there was a `.babelrc` file at the root of the repository, but since I refactor the code and drop this file in order to let `microbundle` manage transpilation with it's own configuration, it looks like I am unable to use `import something from 'some-thing';` anymore in jest tests/mocks...

As `jest` need a `.babelrc` file and I want to let `microbundle` manage babel configuration, having a custom `.babelrc` file will automatically override the configuration.

Running `npm test` always give me this message:
_SyntaxError: Cannot use import statement outside a module_

Here is the **package repository**:
https://github.com/studiomotio/zeroclick/tree/dev

I am probably missing something...

Thanks for your answer :wink: 

> Instead of creating a new issue I am posting a message here as it is probably related.
> If not, let me know and I will open a new issue.

olance:
If you're ending up here from Google, here's an approach that worked for me:

- Create a `babel.config.js` file at the project's root:
```js
// Microbundle uses its own Babel config, so we need to make sure we use this only for tests
[...]

PR #555, opened 2020-01-30
[WIP] add jest transform

therealparmesh:
This addresses https://github.com/developit/microbundle/issues/554.

**NOTE: This needs feedback as it contains structural changes to the project.**

Add something like this to your `package.json` (or wherever your Jest configuration is):

```json
"jest": {
  "transform": {
    ".js": "microbundle/dist/microbundle-jest"
  }
}
```


package.json
```diff
@@ -66,6 +66,7 @@
     "@rollup/plugin-node-resolve": "^6.1.0",
     "asyncro": "^3.0.0",
     "autoprefixer": "^9.7.3",
+    "babel-jest": "^24.8.0",
```
therealparmesh: I moved this to `dependencies` because it is a dependency of code that the package exports (`microbundle-jest.js`).

src/babel-custom.js
```diff
@@ -0,0 +1,160 @@
...
+	return mergedItems;
+};
+
+const createConfigItems = (type, items) => {
+	return items.map(({ name, ...options }) => {
+		return createConfigItem([require.resolve(name), options], { type });
+	});
+};
+
+export const config = (config, { customOptions }) => {
```
therealparmesh: I moved and changed this file, so GitHub won't diff this correctly. I lifted `config` to the module scope and made it exportable.

src/babel-custom.js
```diff
@@ -0,0 +1,160 @@
...
+				Object.assign(
+					merge(
+						{
+							loose: true,
+							useBuiltIns: false,
+							targets: customOptions.targets,
+						},
+						preset.options,
+						{
+							modules: customOptions.jest ? 'auto' : false,
```
therealparmesh: I added a `jest` flag here. I was not sure if I should reuse the `modern` flag here.

src/babel-custom.js
```diff
@@ -0,0 +1,160 @@
...
+			},
+		);
+	} else {
+		babelOptions.presets = createConfigItems('preset', [
+			{
+				name: '@babel/preset-env',
+				targets: customOptions.modern
+					? ESMODULES_TARGET
+					: customOptions.targets,
+				modules: customOptions.jest ? 'auto' : false,
```
therealparmesh: See my above comment.

dakshshah96:
@developit / @ForsakenHarmony: would love to see this land!

marvinhagemeister:
Thank you for your PR 🎉 We believe that `babel-jest` is the better solution. We want to refocus `microbundle` again on what made it great: Being a build tool only and we don't have the manpower to maintain additional areas on top of that.

PR #557, opened 2020-02-02
Implement preset-modules for modern

JoviDeCroock:
`preset-env` is currently in a state where modern isn't correctly eliminating certain polyfills

Link: https://github.com/JoviDeCroock/hooked-form/issues/84

### DIFF
--- a/src/lib/babel-custom.js
+++ b/src/lib/babel-custom.js
@@ -42,6 +42,8 @@ const createConfigItems = (type, items) => {
 	});
 };
 
+const presetEnvRegex = RegExp(/@babel\/(preset-)?env/);
+
 export default babelPlugin.custom(babelCore => {
 	return {
 		// Passed the plugin options.
@@ -93,7 +95,7 @@ export default babelPlugin.custom(babelCore => {
 			const babelOptions = config.options || {};
 
 			const envIdx = (babelOptions.presets || []).findIndex(preset =>
-				preset.file.request.test(/@babel\/(preset-)?env/),
+				presetEnvRegex.test(preset.file.request),
 			);
 
 			const environmentPreset = customOptions.modern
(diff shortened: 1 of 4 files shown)


src/lib/babel-custom.js
```diff
@@ -96,6 +96,10 @@ export default babelPlugin.custom(babelCore => {
 				preset.file.request.includes('@babel/preset-env'),
 			);
 
+			const environmentPreset = customOptions.modern
+				? '@babel/preset-modules'
+				: '@babel/preset-env';
+
```
developit: I think we also need to convert `@babel/preset-env` to use `environmentPreset` on L107:

https://github.com/developit/microbundle/pull/557/files#diff-de034110aceba806ff222e668bb57fc1R107

src/lib/babel-custom.js
```diff
@@ -93,14 +93,18 @@ export default babelPlugin.custom(babelCore => {
 			const babelOptions = config.options || {};
 
 			const envIdx = (babelOptions.presets || []).findIndex(preset =>
-				preset.file.request.includes('@babel/preset-env'),
+				preset.file.request.test(/@babel\/(preset-)?env/),
```
ForsakenHarmony: this seems to be broken

JoviDeCroock: Hmm, I wonder if the previous would've been better

ForsakenHarmony:
now the snapshots seem to be a bit out of order or smth

JoviDeCroock:
I'll run the `--updateSnapshot` again, could it be windows-related?

test/__snapshots__/index.test.js.snap
```diff
@@ -1631,14 +1635,14 @@ modern
 
 
 Build \\"modernLib\\" to dist:
-159 B: modern-lib.modern.js.gz
```
JoviDeCroock: The difference in size is noticeable

developit: (this one mostly comes down to rest parameters being supported in preset-modules but transpiled to arguments access in preset-env)

PR #569, opened 2020-02-17
Add Closure compiler

JoviDeCroock:
Adds the default true option to use the closure compiler and gives the option to opt-out and go for terser instead

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Options
 	--globals        Specify globals dependencies, or 'none'
 	--define         Replace constants with hard-coded values
 	--alias          Map imports to different modules
-	--compress       Compress output using Terser
+	--compress       Compress output using Closure/Terser
 	--strict         Enforce undefined global context and add "use strict"
 	--name           Specify name exposed in UMD builds
 	--cwd            Use an alternative working directory  (default .)
@@ -173,6 +173,7 @@ Options
 	--tsconfig       Specify the path to a custom tsconfig.json
 	--css-modules    Configures .css to be treated as modules (default: null)
 	-h, --help       Displays this message
+	--closure				 Specify if using the closure-compiler, when false microbundle will use terser instead (default: true)
 
 Examples
 	$ microbundle build --globals react=React,jquery=$
--- a/src/index.js
+++ b/src/index.js
@@ -592,7 +592,7 @@ function createConfig(options, entry, format, writeMeta) {
 						},
 					}),
 					options.compress !== false && [
-						options.closure && (modern || format === 'es')
+						options.closure && (format === 'es' || modern)
 							? closureCompiler({ remove_strict_directive: true })
 							: terser({
 									sourcemap: true,
--- a/src/index.js
+++ b/src/index.js
@@ -21,6 +21,7 @@ import json from '@rollup/plugin-json';
 import logError from './log-error';
 import { readFile, isDir, isFile, stdout, stderr, isTruthy } from './utils';
 import camelCase from 'camelcase';
+import closureCompiler from '@ampproject/rollup-plugin-closure-compiler';
 
 const removeScope = name => name.replace(/^@.*\//, '');
 
@@ -590,29 +591,31 @@ function createConfig(options, entry, format, writeMeta) {
 						},
 					}),
 					options.compress !== false && [
-						terser({
-							sourcemap: true,
-							compress: Object.assign(
-								{
-									keep_infinity: true,
-									pure_getters: true,
-									// Ideally we'd just get Terser to respect existing Arrow functions...
-									// unsafe_arrows: true,
-									passes: 10,
-								},
-								minifyOptions.compress || {},
-							),
-							output: {
-								// By default, Terser wraps function arguments in extra parens to trigger eager parsing.
-								// Whether this is a good idea is way too specific to guess, so we optimize for size by default:
-								wrap_func_args: false,
-							},
-							warnings: true,
-							ecma: modern ? 9 : 5,
-							toplevel: modern || format === 'cjs' || format === 'es',
-							mangle: Object.assign({}, minifyOptions.mangle || {}),
-							nameCache,
-						}),
+						options.closure
+							? closureCompiler()
+							: terser({
+									sourcemap: true,
+									compress: Object.assign(
+										{
+											keep_infinity: true,
+											pure_getters: true,
+											// Ideally we'd just get Terser to respect existing Arrow functions...
+											// unsafe_arrows: true,
+											passes: 10,
+										},
+										minifyOptions.compress || {},
+									),
+									output: {
+										// By default, Terser wraps function arguments in extra parens to trigger eager parsing.
+										// Whether this is a good idea is way too specific to guess, so we optimize for size by default:
+										wrap_func_args: false,
+									},
+									warnings: true,
+									ecma: modern ? 9 : 5,
+									toplevel: modern || format === 'cjs' || format === 'es',
+									mangle: Object.assign({}, minifyOptions.mangle || {}),
+									nameCache,
+							  }),
 						nameCache && {
 							// before hook
 							options: loadNameCache,
--- a/src/index.js
+++ b/src/index.js
@@ -592,7 +592,7 @@ function createConfig(options, entry, format, writeMeta) {
 						},
 					}),
 					options.compress !== false && [
-						options.closure
+						options.closure && (modern || format === 'es')
 							? closureCompiler()
 							: terser({
 									sourcemap: true,
--- a/src/index.js
+++ b/src/index.js
@@ -556,6 +556,7 @@ function createConfig(options, entry, format, writeMeta) {
 								},
 							},
 							tsconfig: options.tsconfig,
+							objectHashIgnoreUnknownHack: !!options.closure,
 							tsconfigOverride: {
 								compilerOptions: {
 									target: 'esnext',
--- a/src/index.js
+++ b/src/index.js
@@ -593,7 +593,7 @@ function createConfig(options, entry, format, writeMeta) {
 					}),
 					options.compress !== false && [
 						options.closure && (modern || format === 'es')
-							? closureCompiler()
+							? closureCompiler({ remove_strict_directive: true })
 							: terser({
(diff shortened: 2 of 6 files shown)


src/index.js
```diff
@@ -555,6 +556,7 @@ function createConfig(options, entry, format, writeMeta) {
 								},
 							},
 							tsconfig: options.tsconfig,
+							objectHashIgnoreUnknownHack: !!options.closure,
```
ForsakenHarmony: what's that?

src/prog.js
```diff
@@ -37,6 +37,11 @@ export default handler => {
 		.option('--name', 'Specify name exposed in UMD builds')
 		.option('--cwd', 'Use an alternative working directory', '.')
 		.option('--sourcemap', 'Generate source map', true)
+		.option(
+			'--closure',
+			'Use the closure-compiler for minifying instead of terser',
+			true,
```
ForsakenHarmony: should this be the default?

ForsakenHarmony:
Also does this significantly increase build times?

test/__snapshots__/index.test.js.snap
```diff
@@ -1176,7 +1177,7 @@ exports[`fixtures build css-modules--string with microbundle 3`] = `
 `;
 
 exports[`fixtures build css-modules--string with microbundle 4`] = `
-"export default function(){var a=document.createElement(\\"div\\");return a.className=\\"_contains_this_0a8c24df242c2cd708036873307aea94 _contains_this_81567d0efc15a456670452d3277e1a68\\",a}
+"var b={test_class_that_should_be_scoped:\\"_contains_this_81567d0efc15a456670452d3277e1a68\\"},c={scoped_class:\\"_contains_this_0a8c24df242c2cd708036873307aea94\\"};export default function(){var a=document.createElement(\\"div\\");a.className=c.scoped_class+\\" \\"+b.test_class_that_should_be_scoped;return a}
```
ForsakenHarmony: these are a bit disappointing I guess

src/index.js
JoviDeCroock: Without it async functions won't work with the ts plugin when combined with closure

https://github.com/ezolenko/rollup-plugin-typescript2/issues/105

src/prog.js
JoviDeCroock: It does seem equally as fast as terser from what I'm seeing

JoviDeCroock:
@ForsakenHarmony  It does seem equally as fast as terser from what I'm seeing

test/__snapshots__/index.test.js.snap
```diff
@@ -1642,15 +1649,15 @@ Build \\"mangleJsonFile\\" to dist:
...
+84 B: mangle-json-file.esm.js.br
 198 B: mangle-json-file.umd.js.gz
 164 B: mangle-json-file.umd.js.br"
 `;
 
 exports[`fixtures build mangle-json-file with microbundle 2`] = `6`;
 
 exports[`fixtures build mangle-json-file with microbundle 3`] = `
-"var o={prop1:1,__p2:2};export default function(){return console.log(o.prop1),console.log(o.__p2),o}
+"var a={prop1:1,_prop2:2};export default function(){console.log(a.prop1);console.log(a._prop2);return a}
```
kristoferbaxter: This looks like a bug or issue to resolve before merging, the `mangled` value isn't being used.

README.md
```diff
@@ -173,6 +173,7 @@ Options
 	--tsconfig       Specify the path to a custom tsconfig.json
 	--css-modules    Configures .css to be treated as modules (default: null)
 	-h, --help       Displays this message
+	--closure				 Specify if using the closure-compiler, when false microbundle will use terser instead (default: true)
```
developit: If the default is to use closure, the option should be `--terser`. See other comment for discussion of whether default is correct here.

src/index.js
```diff
@@ -590,29 +592,31 @@ function createConfig(options, entry, format, writeMeta) {
...
-								// Whether this is a good idea is way too specific to guess, so we optimize for size by default:
-								wrap_func_args: false,
-							},
-							warnings: true,
-							ecma: modern ? 9 : 5,
-							toplevel: modern || format === 'cjs' || format === 'es',
-							mangle: Object.assign({}, minifyOptions.mangle || {}),
-							nameCache,
-						}),
+						options.closure && (format === 'es' || modern)
```
developit: We should clarify in the docs/help that Terser is always used for `cjs,iife,umd` formats, regardless of the `--closure` option.

src/prog.js
developit: Since this is a tool people universally expect to pass options to (due to no config), I think we should start out by merging this with the default still as Terser. We can quickly cut over to Closure as a new default once it's sat on master for a version cycle. I just don't want landing this PR and getting closure support published to necessitate a major version change all at once.

test/__snapshots__/index.test.js.snap
```diff
@@ -20,16 +20,16 @@ alias
 Build \\"aliasMapping\\" to dist:
 62 B: alias-mapping.js.gz
 46 B: alias-mapping.js.br
-62 B: alias-mapping.esm.js.gz
-46 B: alias-mapping.esm.js.br
+61 B: alias-mapping.esm.js.gz
```
developit: as per side-discussion, this size decrease is due to custom property name mappings being ignored.

test/__snapshots__/index.test.js.snap
```diff
@@ -968,7 +968,8 @@ Build \\"classDecoratorsTs\\" to dist:
 exports[`fixtures build class-decorators-ts with microbundle 2`] = `7`;
 
 exports[`fixtures build class-decorators-ts with microbundle 3`] = `
-"var e=function(){function e(e){this.greeting=e}return e.prototype.greet=function(){return\\"Hello, \\"+this.greeting},e}(),t=new(e=function(e,t,r,n){var o,c=arguments.length,l=c<3?t:null===n?n=Object.getOwnPropertyDescriptor(t,r):n;if(\\"object\\"==typeof Reflect&&\\"function\\"==typeof Reflect.decorate)l=Reflect.decorate(e,t,r,n);else for(var f=e.length-1;f>=0;f--)(o=e[f])&&(l=(c<3?o(l):c>3?o(t,r,l):o(t,r))||l);return c>3&&l&&Object.defineProperty(t,r,l),l}([function(e){Object.seal(e),Object.seal(e.prototype)}],e))(\\"Hello World\\");export default t;
+"var h=function(){function a(a){this.greeting=a}a.prototype.greet=function(){return\\"Hello, \\"+this.greeting};return a}();
```
developit: newline here seems odd...

test/__snapshots__/index.test.js.snap
developit: It's an object coming from an import - I'm thinking closure can't inline because its crossing a module boundary?

@kristoferbaxter maybe cross-module code motion is only enabled for ADVANCED_OPTIMIZATIONS?

test/__snapshots__/index.test.js.snap
```diff
@@ -1403,16 +1404,16 @@ default-named
...
+55 B: default-named.esm.js.br
 170 B: default-named.umd.js.gz
 124 B: default-named.umd.js.br"
 `;
 
 exports[`fixtures build default-named with microbundle 2`] = `6`;
 
 exports[`fixtures build default-named with microbundle 3`] = `
-"var t=42;export default function(){}export{t as foo};
+"export default function(){};var foo=42;export{foo}
```
developit: This is a fantastic output.

test/__snapshots__/index.test.js.snap
```diff
@@ -1493,16 +1494,22 @@ esnext-ts
...
+"function h(a){if(\\"undefined\\"!==typeof Symbol){if(Symbol.asyncIterator){var b=a[Symbol.asyncIterator];if(null!=b)return b.call(a)}if(Symbol.iterator&&(b=a[Symbol.iterator],null!=b))return b.call(a)}throw new TypeError(\\"Object is not async iterable\\");}
```
developit: Is there a default line length configuration in Closure or something? It's strange there would be line breaks in the output at all.

test/__snapshots__/index.test.js.snap
developit: https://github.com/developit/microbundle/wiki/mangle.json#for-manglejson-implementors

src/index.js
agilgur5: Just randomly saw this linked, but per the last comment there (from me), it's no longer necessary if you update to v0.26.0+ .

And per the other comments / former docs, if you are using the `objectHashIgnoreUnknownHack` option, you should really use it with `clean: true` as otherwise it can cause caching issues.

src/prog.js
kristoferbaxter: That’s a fair compromise. Let’s make this an option (not default) to land, and iterate on output.

To become the default I think we need to support the custom mangle format microbundle uses (uglify’s native format). 

Also, the newlines should be addressed for consistency. (This work should likely be part of the rollup closure plugin).

test/__snapshots__/index.test.js.snap
```diff
@@ -1225,7 +1226,7 @@ exports[`fixtures build css-modules--true with microbundle 3`] = `
 `;
 
 exports[`fixtures build css-modules--true with microbundle 4`] = `
-"export default function(){var e=document.createElement(\\"div\\");return e.className=\\"_2kWDE _1E6DU\\",e}
+"var b={test_class_that_should_be_scoped:\\"_1E6DU\\"},c={scoped_class:\\"_2kWDE\\"};export default function(){var a=document.createElement(\\"div\\");a.className=c.scoped_class+\\" \\"+b.test_class_that_should_be_scoped;return a}
```
JoviDeCroock: @kristoferbaxter this one is quite the downgrade, I'll file an issue for this

test/__snapshots__/index.test.js.snap
kristoferbaxter: Correct @developit

ForsakenHarmony:
we can merge this now with terser still being the default and decide later if we want to make closure the default

PR #570, opened 2020-02-17
Fix babel cache bleed

### LINKED ISSUE #568: Specifying `commonjs` as one of the output formats forces generators to compile to regenerator runtime
`--format modern,es,umd`

generators are left untouched

`--format modern,es,umd,cjs`

transpiles generators for all formats, no matter where `cjs` is inserted.

Would be happy to make a fix if you can point me in the right direction.

developit:
Instantiating babel.custom() separately for each output should fix #568, which I believe is caused by cached configuration bleeding from `cjs` into other output targets.


src/index.js
```diff
@@ -467,7 +467,7 @@ function createConfig(options, entry, format, writeMeta) {
 
 	const externalPredicate = new RegExp(`^(${external.join('|')})($|/)`);
 	const externalTest =
-		external.length === 0 ? () => false : id => externalPredicate.test(id);
+		external.length === 0 ? id => false : id => externalPredicate.test(id);
```
ForsakenHarmony: does this affect anything?

src/index.js
```diff
@@ -630,9 +631,11 @@ function createConfig(options, entry, format, writeMeta) {
...
 							config._sizeInfo = Promise.all(
-								Object.values(bundle).map(({ code, fileName }) =>
-									code ? getSizeInfo(code, fileName, options.raw) : false,
-								),
+								Object.values(bundle).map(({ code, fileName }) => {
+									if (code) {
+										return getSizeInfo(code, fileName, options.raw);
+									}
+								}),
 							).then(results => results.filter(Boolean).join('\n'));
```
ForsakenHarmony: I guess the implicit undefined return will work :thinking:

src/index.js
```diff
@@ -622,6 +622,7 @@ function createConfig(options, entry, format, writeMeta) {
 									fs.writeFile(
 										getNameCachePath(),
 										JSON.stringify(nameCache, null, 2),
+										() => {},
```
ForsakenHarmony: same as above, does this change the behavior?

src/index.js
developit: no, just typescript was yelling at me about it.

src/index.js
developit: same deal, was a typescript warning

src/index.js
developit: yeah cause we cast to boolean on L639. this was another TS warning lol

ForsakenHarmony:
bors retry

ForsakenHarmony:
bors retry

ForsakenHarmony:
bors retry

PR #571, opened 2020-02-28
Support import JSX from JS file

n3tr:
## Changes

- Support import JSX from JS file by adding JSX extension to `@rollup/plugin-node-resolve` plugin 

## Background

I have experience an error when importing `.jsx file` from `.js` file like

```
index.js -> A.jsx
```

will cause an error  

```
Could not resolve './A' from src/index.js
```

However, Importing `.jsx` file from TypeScript file (`.ts`) file is seems working fine except the the import is chaining more than 3 levels depth. 

```sh
index.ts -> A.jsx # This is fine
index.ts -> A.jsx  -> B.jsx # This is fine

index.ts -> A.jsx  -> B.jsx -> C.jsx 
# Error: Could not resolve './C' from src/jsx-components/B.jsx
```

To reproduce the issue

```sh
git clone -b import-jsx-error --single-branch git@github.com:n3tr/microbundle-example.git
cd microbundle-example

yarn

yarn build --jsx React.createElement ./src/index.js 
# or
yarn build --jsx React.createElement ./src/index.ts

```

---

I'm not really sure what is the root cause of the error, by adding extensions seems to fix the problem.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -527,6 +527,7 @@ function createConfig(options, entry, format, writeMeta) {
 					nodeResolve({
 						mainFields: ['module', 'jsnext', 'main'],
 						browser: options.target !== 'node',
+						extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
 					}),
 					commonjs({
 						// use a regex to make sure to include eventual hoisted packages
--- a/src/index.js
+++ b/src/index.js
@@ -527,6 +527,7 @@ function createConfig(options, entry, format, writeMeta) {
 					nodeResolve({
 						mainFields: ['module', 'jsnext', 'main'],
 						browser: options.target !== 'node',
+						// defaults + .jsx
 						extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -527,6 +527,7 @@ function createConfig(options, entry, format, writeMeta) {
 					nodeResolve({
 						mainFields: ['module', 'jsnext', 'main'],
 						browser: options.target !== 'node',
+						extensions: ['.mjs', '.js', '.jsx', '.json', '.node'],
```
ForsakenHarmony: just kinda wondering if there could be problems with typescript, probably not because everything in node modules should already be compiled

n3tr: I just add `jsx` to the list of [the default values](https://github.com/rollup/plugins/blob/a66c8c70fc18f3967783b4e3e16e0aecd437d1f1/packages/node-resolve/src/index.js#L25), nothing should be different for TypeScript.
:)

ForsakenHarmony:
Thanks

PR #588, opened 2020-04-16
Re-add support for --define @expression=replacement

developit:
This should match Terser's [`global_defs` option](https://github.com/terser/terser/#conditional-compilation-api).

/cc @jviide


jviide:
LGTM 👍

PR #589, opened 2020-04-17
Add --generate-types option

developit:
This requires TS 3.7: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#--declaration-and---allowjs

One thing I wasn't sure of: maybe this could just be enabled by default? Currently I think it's possible via `tsconfig.json`, but doing so would potentially run TypeScript on all of node_modules.

TODO:

- [ ] tests


marvinhagemeister:
+1 on enabling it as default

Hotell:
> maybe this could just be enabled by default?

Absolutely (even folks not aware about typescript will ship at least some amount of definitions - happy consumers)

> doing so would potentially run TypeScript on all of node_modules.

It would not:  setting `"skipLibCheck": true,` flag will mitigate that (it's turned on by default since TS 3.9)

src/index.js
```diff
@@ -571,6 +574,10 @@ function createConfig(options, entry, format, writeMeta) {
 									target: 'esnext',
 								},
 							},
+							include: ['*.ts+(|x)', '**/*.ts+(|x)'].concat(
+								emitDeclaration ? ['*.js+(|x)', '**/*.js+(|x)'] : [],
+							),
+							exclude: ['**/node_modules/**', '*.d.ts', '**/*.d.ts'],
```
Hotell, line 580: node_modules are excluded by default

src/index.js
```diff
@@ -571,6 +574,10 @@ function createConfig(options, entry, format, writeMeta) {
 									target: 'esnext',
 								},
 							},
+							include: ['*.ts+(|x)', '**/*.ts+(|x)'].concat(
```
Hotell, line 577: not needed:
- default TS adds `*.ts,.*tsx` to the glob
- setting `allowJs` adds `*.js,*.jsx` to the glob

src/index.js
```diff
@@ -472,6 +472,7 @@ function createConfig(options, entry, format, writeMeta) {
 			: () => resolve(options.cwd, 'mangle.json');
 
 	const useTypescript = extname(entry) === '.ts' || extname(entry) === '.tsx';
+	const emitDeclaration = !!options.generateTypes;
```
Hotell, line 475: maybe just checking if `pkg.types || pkg.typings` is present in package.json would be enough to trigger this?

marvinhagemeister: Love that idea! Made https://github.com/developit/microbundle/pull/631 that does enable declarations only when `pkg.types || pkg.typings` is present 🎉

developit: @marvinhagemeister I copied your change into this PR 👍

src/index.js
```diff
@@ -498,6 +499,8 @@ function createConfig(options, entry, format, writeMeta) {
 								compilerOptions: {
 									sourceMap: options.sourcemap,
 									declaration: true,
+									allowJs: useTypescript || emitDeclaration,
```
katywings, line 502: Since the same check already happens on line 490, allowJs is always true here, personally I would replace it with `allowJs: true` 😁

katywings (approved):
Looking good ^.^

src/index.js
developit: Ha! Good catch.

developit:
```suggestion
									allowJs: true,
```

marvinhagemeister (approved):

PR #624, opened 2020-05-20
refactor: split utilities in separate files

developit:
Extracted multiple utilities out of the main isFile

/utils.js
- isDir
- isFile
- stdout
- stderr
- isTruthy
- removeScope

/lib/compressed-size.js
- getSizeInfo

/lib/terser.js
- normalizeMinifyOptions

/lib-option-normalization.js
-	parseAliasArgument,
- parseMappingArgument
- toReplacementExpression

/lib/package-info
- getConfigFromPkgJson
- getName

/lib/css-modules
- shouldCssModules
- cssModulesConfig

### DIFF
--- a/src/lib/css-modules.js
+++ b/src/lib/css-modules.js
@@ -1,0 +1,49 @@
+export function shouldCssModules(options) {
+	const passedInOption = processCssmodulesArgument(options);
+
+	// We should module when my-file.module.css or my-file.css
+	const moduleAllCss = passedInOption === true;
+
+	// We should module when my-file.module.css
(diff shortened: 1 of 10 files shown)


src/lib/css-modules.js
```diff
@@ -0,0 +1,49 @@
...
+		}
+
+		return { generateScopedName };
+	}
+
+	return false;
+}
+
+/**
+ * This is done becuase if you use the cli default property, you get a primiatve "null" or "false",
```
marvinhagemeister, line 36: Nit: Typo `becuase` -> `because`

PR #630, opened 2020-05-25
Clear generated types before test run

marvinhagemeister:
- Clear generated type folder before running each test

~~Just cleared my `node_modules` and reinstalled them. From the looks of it the newer dependencies than what I had locally lead to some snapshot changes.~~

(_Turns out that this was caused by a different node version_)


ForsakenHarmony:
I'm guessing you're trying to solve some problem you ran into?

marvinhagemeister:
Yeah, currently only the `dist` directory is cleared before each test run, but not `types`.  If a previous test run was successful and something broke in a new commit the CI will pass either way because of the old files.

PR #631, opened 2020-05-25
Only generate types when requested

marvinhagemeister:
This PR changes the default behaviour to only generate types when the `types` or `typings` field was found in `package.json`.

Replaces #589 .

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -566,7 +566,7 @@ function createConfig(options, entry, format, writeMeta) {
 							tsconfigDefaults: {
 								compilerOptions: {
 									sourceMap: options.sourcemap,
-									declaration: true,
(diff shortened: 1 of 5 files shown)


ForsakenHarmony:
This seems to kinda be the inverse of #589?

developit:
I think this needs the [`emitDeclaration` tweak](https://github.com/developit/microbundle/pull/589/files#diff-1fdf421c05c1140f6d71444ea2b27638R557) from #589 - currently if there are no `.ts` files, the typescript rollup plugin doesn't get run. Also the allowJs bit (though that could perhaps be on-by-default).

marvinhagemeister:
True, looks like I misunderstood the original PR 🙈

developit:
lol not at all, I'll try to make this change there

PR #636, opened 2020-05-29
Update option docs for --externals (default peerDependencies & dependencies)

gtrufitt:
Just a small clarification that the externals default is `peerDependencies` and `dependencies`.

(Not sure if this is *needed* as obviously this is a sensible default, but it wasn't clear until checking the code that we didn't need to have `external`s that mirrored `peerDependencies` like you might if setting up rollup from scratch)

https://github.com/developit/microbundle/blob/ded90ace965eabdf1707d1a75b20629513086775/src/index.js#L406

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ Options
 	-f, --format     Only build specified formats (any of modern,es,cjs,umd or iife) (default modern,es,cjs,umd)
 	-w, --watch      Rebuilds on any change  (default false)
 	--target         Specify your target environment (node or web)  (default web)
-	--external       Specify external dependencies, or 'none'
+	--external       Specify external dependencies, or 'none' (default peerDependencies from package.json)
 	--globals        Specify globals dependencies, or 'none'
 	--define         Replace constants with hard-coded values
 	--alias          Map imports to different modules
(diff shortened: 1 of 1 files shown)


gtrufitt:
Thanks @developit 

So it looks like the default is some stdLib stuff, `peerDep` and `dep`s.

If the option.external is set then it's the (comma separated) external entries, stdLib and *just* `peerDep`s? 

The `options.entries` being the other stickler, which maybe is not important to call out in the default in docs?

Hence, my assumption for a minimal but useful `default`, if external isn't set, in docs was 'peerDependencies and Dependencies' - but happy to push a change if you think there's more needed?

```js

let external = ['dns', 'fs', 'path', 'url'].concat(
		options.entries.filter(e => e !== entry),
	);

const peerDeps = Object.keys(pkg.peerDependencies || {});
	if (options.external === 'none') {
		// bundle everything (external=[])
	} else if (options.external) {
		external = external.concat(peerDeps).concat(options.external.split(','));
	} else {
		external = external
			.concat(peerDeps)
			.concat(Object.keys(pkg.dependencies || {}));
	}
```

wardpeet:
This is great! We might want to show while outputting in verbose mode or something which packages are being externalized but for this PR it looks 👍

gtrufitt:
👍

krobing:
I think the externals added as other more, should be in addition to `deps/peerDeps`.

> My only nit would be that I think modules passed via --external are marked as external in _addition_ to deps/peerDeps, rather than those being an overridable default (I may be misremembering).

Issue #643, opened 2020-05-31
Wrong typescript emit path if --cwd is used


katywings:
IS: If you use the cli cwd option to build a nested package `xyz`, the types are emitted into `./dist/`.

SHOULD: Types should be emitted into the dist folder `xyz/dist` of the nested package.

DEMO: 
- Clone https://github.com/katywings/microbundle-test/tree/65ba5c4aa436a9807969dbff9dcab30a9226cfc2 
- `npm install && ./scripts/build-devtools.js`
- The types will be emitted into `dist/devtools/lib/main.d.ts` instead of `devtools/dist/main.d.ts`

REMARKS:
- If the devtools/lib/main.ts doesnt import something from an external, the ts emit path is different but still wrong: `dist/main.d.ts`
- Not directly related to the cwd thing, but I noticed that the `--external` cli option "freaks out" if the external has a dash character: `microbundle` prints: `No name was provided for external module 'microbundle-test' in output.globals – guessing 'microbundleTest'`. Doing the exact same bundle but naming the global "microbundletest" doesnt print this warning. (P.S. if you try this you also have to change the tsconfig.json path alias from microbundle-test to microbundletest)

developit:
Ahh good point about external - we should be pre-populating the name when generating entry point externals but we aren't!

developit:
Ah I found the cwd issue - we [pass the package.json `types` path directly](https://github.com/developit/microbundle/blob/58288ec989d0a5dbd6ed500481918cf0ced39c66/src/index.js#L571) to rollup-plugin-typescript2, but should be resolving it against the cwd option value.

katywings:
- P.S. i can't really make a fix regarding https://github.com/developit/microbundle/issues/643#issuecomment-636535838 since I don't yet have the whole picture on how the entry point externals work together with different configurations.

[...]

PR #646, opened 2020-06-02
fix: set types declarationDir based on cwd (#643)

katywings:
This includes the change regarding: https://github.com/developit/microbundle/issues/643#issuecomment-636536332

I tested it with `npm link` on my https://github.com/katywings/microbundle-test repository.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -364,6 +364,21 @@ async function getOutput({ cwd, output, pkgMain, pkgName }) {
 	return main;
 }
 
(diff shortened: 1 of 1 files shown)


developit:
(sidebar: this makes me wonder if there's a way to set the default declaration filename too)

marvinhagemeister:
@developit TS unfortunately has no option for that. It picks the filenames based on the original source files.

PR #649, opened 2020-06-05
Fix a mistake in the readme

AlexHayton:


developit:
This is backwards now - "esmodule" is the modern bundle.

wardpeet:
You're right I'll recert

PR #650, opened 2020-06-05
fix: escape rollup external test items

katywings:
During some investigation in the rollup external test logic I noticed, that the external parts are joined into a regex without any escaping. 

Background:
Originally I thought this was the reason for some problems with bundling "lodash.isPlainObject" (notice the dot) but that actually has some other cause... Still I think that adding the regex escaping might be wise here. Let me know what you think ;).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2534,6 +2534,13 @@
         "ansi-styles": "^3.2.1",
         "escape-string-regexp": "^1.0.5",
         "supports-color": "^5.3.0"
+      },
+      "dependencies": {
+        "escape-string-regexp": {
+          "version": "1.0.5",
+          "resolved": "https://registry.npmjs.org/escape-string-regexp/-/escape-string-regexp-1.0.5.tgz",
+          "integrity": "sha1-G2HAViGQqN/2rjuyzwIAyhMLhtQ="
+        }
       }
     },
     "chardet": {
@@ -2984,6 +2991,11 @@
             "supports-color": "^2.0.0"
           },
           "dependencies": {
+            "escape-string-regexp": {
+              "version": "1.0.5",
+              "resolved": "https://registry.npmjs.org/escape-string-regexp/-/escape-string-regexp-1.0.5.tgz",
+              "integrity": "sha1-G2HAViGQqN/2rjuyzwIAyhMLhtQ="
+            },
             "supports-color": {
               "version": "2.0.0",
               "resolved": "https://registry.npmjs.org/supports-color/-/supports-color-2.0.0.tgz",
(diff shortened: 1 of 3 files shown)


ForsakenHarmony:
Hmm, there could be some use for passing a regex to it as well though

katywings:
@ForsakenHarmony 🤔 hmm, an alternative could be to only escape the peerDeps which are concated into `external` somewhere around here: https://github.com/developit/microbundle/blob/master/src/index.js#L415

developit:
@katywings that seems like a good compromise.

katywings:
@developit / @ForsakenHarmony I wanted to make the necessary changes and add the escaping only to externals originating from pkg.deps. But after spending some time on this I saw this line:
https://github.com/developit/microbundle/blob/master/src/index.js#L429
Microbundle actually sets up globals analog to the entries from `external`. 

Unless we gonna change this as well, It doesn't make sense to support regex patterns in the externals since this would screw with the globals. With this new knowledge in mind, my recommendation to you is to add the external escaping analog to this PR as is and then at a later point think about how a regex support in external options could look like - including thinking about sideeffects regarding globals etc.

krobing:
Please don't take this feature out, I'm using `RegExp`S

katywings:
@krobing: where do you use regex? In the `--external` cli argument? Can you give us an example?

krobing:
Yeah, I've been using it this way:

```js
"scripts": {
  "bundle": "microbundle --jsx React.createElement --no-compress --external ../../public/bundle.js,[\\w@$]+[\\w-/]*(\\.[\\w]+)? --alias styled-components=styled-components/macro" 
}
```

I notice that in [#636 ](https://github.com/developit/microbundle/pull/636) they solves it by default adding the `deps/peerDeps` to external modules, but when I want to add other more paths in additon with them, the `deps` are no longer marked as external any more.

katywings:
@krobing: thanks for your details! Is the goal of your regex to keep all src files external, or whats the end goal? (just so that we have the full picture ;)

katywings:
@marvinhagemeister / @developit: if okay from you, I'll make a pr where I remove the escaping again and add a code comment/readme info that this is an intended feature. What do you think about adding a flag like "--external-regex" which is enabled by default?

developit:
I am wondering what the level of effort would be to do the escapement for generated and keep regex supported.

katywings:
@developit:
i think it like this:
- pkg deps -> escaped in externals
- pkg deps -> escaped in globals
- external cli arg -> escaped in globals
- external cli arg -> not escaped in externals 

I probably was like half way going for this when I originally planned out my earlier fix 😅

developit:
Heh yeah I think so. I just opened #665 that I believe should be exactly what you described (good list!).

krobing:
https://github.com/developit/microbundle/pull/650#issuecomment-646762209 it's not necessarily to keep src files as external, in my case I just want to refer "../../public/bundle.js" as an external module, but not the src files. thanks to you.

PR #658, opened 2020-06-12
feat: add --pkg-main option

katywings:
The goal of this feature is to add a solution for microbundle, to manually disable the package.json / format suffix automations analog to https://twitter.com/Katy_Wings/status/1271189844077748224

The new option `--pkg-main` by default is true - basically notifying the consumer that main entries in package.json will be considered. The consumer can supply `--pkg-main false` to disable the microbundle package.json "sideffects".

This PR also includes a little refactor of the "main" generation - most of it is the same as before but just moved into separate functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,14 @@ Microbundle uses the fields from your `package.json` to figure out where it shou
 }
 ```
 
+### Building a single bundle with a fixed output name
+
+By default Microbundle outputs multiple bundles, one bundle per format. A single bundle with a fixed output name can be built like this:
+
+```bash
+microbundle -i lib/main.js -o dist/bundle.js --no-pkg-main -f umd
+```
+
 ### Mangling Properties
 
 To achieve the smallest possible bundle size, libraries often wish to rename internal object properties or class members to smaller names - transforming `this._internalIdValue` to `this._i`. Microbundle doesn't do this by default, however it can be enabled by creating a `mangle.json` file (or a `"mangle"` property in your package.json). Within that file, you can specify a regular expression pattern to control which properties should be mangled. For example: to mangle all property names beginning an underscore:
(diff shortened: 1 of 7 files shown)


developit:
Nice - @wardpeet and I were looking at breaking those functions out anyway, and I like the way you've done it here!

README.md
```diff
@@ -178,6 +178,14 @@ Microbundle uses the fields from your `package.json` to figure out where it shou
 }
 ```
 
+### Building a single bundle with a fixed output name
+
+By default Microbundle outputs multiple bundles, one bundle per format. A single bundle with a fixed output name can be built like this:
+
+```bash
+microbundle -i lib/main.js -o dist/bundle.js --pkg-main false -f umd
```
developit, line 186: tbh I'm not sure which is clearer, but I thought I'd mention that the `no-` prefix is also supported:

```suggestion
microbundle -i lib/main.js -o dist/bundle.js --no-pkg-main -f umd
```

src/index.js
```diff
@@ -392,6 +392,54 @@ async function getEntries({ input, cwd }) {
...
+		dirname(filename),
+		name + basename(filename).replace(/^[^.]+/, ''),
+	);
+}
+
+function getMain({ options, entry, format }) {
+	const { pkg } = options;
+	const pkgMain = options['pkg-main'];
+
+	if (!pkgMain) {
```
developit, line 406: I wonder if we could even apply this same logic if the value provided for the `--output` argument has a file extension. Like `microbundle -f umd -o bundle.js` would see `.js` and know to use that name as-is.

README.md
katywings: We could change it, since it would have the advantage of implicitly documenting that `--no-...` also exists 😅

src/index.js
katywings: I think this would open to questions:
- Such a new logic could be a "breaking" change because there might be people already using -o with ".js" and expecting it to add the format suffix. If a breaking change would mean that this has to wait for 1.0 I personally would prefer to split this idea into another issue 😇
- If we implement this logic, it should only ever apply if the format option is set with a single value

marvinhagemeister: Agree with @katywings here. I'm worried that us making too many assumptions will lead to unexpected results for the user. In either case this would be a breaking change.

PR #626, opened 2020-05-22
Fix auto installing buggy @rollup/plugin-commonjs - v1.11.0

mihilranathunga:
resolves rollup/plugins#304

PR #660, opened 2020-06-15
Update dependencies

developit:
This updates most things, but importantly `@rollup/plugin-commonjs` to `13.x`, which fixes #626.

The snapshot updates are all because Autoprefixer correctly now stops injecting `display:-webkit-flex` for the default browserslist - it should never have been injecting it.


jakewies:
I hate to pry, but is there a time frame for when this will be released? Just trying to plan my day 👍

PR #663, opened 2020-06-18
add support for amd options

hanai:
add support for amd options

https://rollupjs.org/guide/en/#outputamd

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,8 @@ Options
 	--compress       Compress output using Terser
 	--strict         Enforce undefined global context and add "use strict"
 	--name           Specify name exposed in UMD and IIFE builds
+	--amd.id         ID for AMD module (default is anonymous)
+	--amd.define     Function to use in place of `define`
 	--cwd            Use an alternative working directory  (default .)
 	--sourcemap      Generate source map  (default true)
 	--raw            Show raw byte size  (default false)
--- a/src/index.js
+++ b/src/index.js
@@ -457,6 +457,15 @@ function createConfig(options, entry, format, writeMeta) {
 		outputAliases['.'] = './' + basename(options.output);
 	}
 
+	let amd = {};
+	if (options['amd.id']) {
+		amd.id = options['amd.id'];
+	}
+
+	if (options['amd.define']) {
+		amd.define = options['amd.define'];
+	}
+
 	const moduleAliases = options.alias
 		? parseMappingArgumentAlias(options.alias)
 		: [];
@@ -707,6 +716,7 @@ function createConfig(options, entry, format, writeMeta) {
 		outputOptions: {
 			paths: outputAliases,
 			globals,
+			amd,
 			strict: options.strict === true,
 			legacy: true,
 			freeze: false,


developit:
I'm a little worried about these AMD-specific parameters.

Out of curiosity - did you try using `--name`? It should cause AMD output to use named define instead of anonymous. The `amd.define` option seems okay, though I'd be tempted to omit it from the documentation and just support it as a hidden feature.

developit:
As a thought, what about having an `--amdNamed` boolean option that causes `amd.id` to be set to `--name`?

README.md
```diff
@@ -229,6 +229,8 @@ Options
 	--compress       Compress output using Terser
 	--strict         Enforce undefined global context and add "use strict"
 	--name           Specify name exposed in UMD and IIFE builds
+	--amd.id         ID for AMD module (default is anonymous)
+	--amd.define     Function to use in place of `define`
```
developit, line 233:
```suggestion
	--amd-name       [amd] use named define("name", ...)
	--amd-define     [amd] function to use in place of define()
```

src/index.js
```diff
@@ -457,6 +457,15 @@ function createConfig(options, entry, format, writeMeta) {
...
 	}
 
+	let amd = {};
+	if (options['amd.id']) {
+		amd.id = options['amd.id'];
+	}
+
+	if (options['amd.define']) {
+		amd.define = options['amd.define'];
+	}
```
developit, line 467:
```suggestion
	let amd = {};
	if (options.amdName) {
		amd.id = options.amdName;
	}

	if (options.amdDefine) {
		amd.define = options.amdDefine;
	}
```

developit:
If possible, I'd love to get a better understanding of why named define is useful. I've written a few AMD loaders in the past and only ever needed named defines or custom define functions as a workaround.

hanai:
> If possible, I'd love to get a better understanding of why named define is useful. I've written a few AMD loaders in the past and only ever needed named defines or custom define functions as a workaround.

In case of `define` is not named as `define` in global, such as `fmd.define`.

wardpeet:
@hanai We feel that adding this feature adds a bit more complex than we want to microbundle. For your use-case, it would make more sense to go with rollup directly instead of microbundle.

We do see value in adding support for creating named amd modules. If you could refactor this PR to support the `amdNamed` boolean, that would be great.

Thanks for understanding and contributing to microbundle!

rart:
@wardpeet @developit I'm running into this issue too — i.e. the lack of support for the amd rollup options. I was thinking submitting a PR and found this one.

The `amdNamed` option only half-solves `amd.id`. Rollup has those props separate, one is the name/namespace for the module and the other — the `amd.id` — is used on the define call to define the module with an id.

The amd.id is pretty important since there are certain module loaders that don't support anonymous definitions. Also, when writing a custom module loader, the id that comes to the define function is useful in countless ways — up to the lib developer. 

Supporting these options would be suuuuper simple — could be as simple as adding
```js
outputOptions: {
  ...
  // if options.amdId or options.amdDefine is undefined rollup takes it as not supplied and behaves as default
  amd: {
    id: options.amdId,
    define: options.amdDefine
  },
  ...
}
```

> @developit: If possible, I'd love to get a better understanding of why named define is useful. I've written a few AMD loaders in the past and only ever needed named defines or custom define functions as a workaround.

As an example, you may have the define function under a certain namespace (e.g. myCompany.define) so amd.define = myCompany.define would give you the right output, otherwise it will always call define on the global object and fail if it isn't there.

I'm happy to submit a PR once I hear back from you guys that it would be accepted. Please suggest approach too to minimize PR revision back-and-forth.

bensalilijames:
Would be keen to see this! Running into this when packaging a library for use in Decentraland, which requires named AMD modules with the ID being the same as the NPM package name.

FWIW, I'd also be keen to have the flexibility to separately define the `amd.id` and `amd.define` options as in the PR currently, instead of a `--amdNamed` boolean. In my case, I want to have a UMD build with the browser global having a different name from the AMD ID. If they're separate options in the CLI, I then have the power to do so!

Understand that this increases the surface area of microbundle, though I think it's a worthwhile addition that unlocks an edge case, without requiring authors to eject from microbundle and roll their own rollup config.

Would you be open to this PR as-is if I resolved the merge conflicts? What do you think? Thank you! 🙌

othree:
I have the same issue. If the module is not named. There will be an issue working with the require.js. My use case is we are writing 3rd party widget. Then use our script on a site are using require.js will see this error.

Reference: https://requirejs.org/docs/errors.html#mismatch

PR #665, opened 2020-06-19
Allow RegExp --externals only for the CLI argument

developit:
This is an addendum to #650. It escapes externals that are generated from package.json or static lists, while still allowing regular expression `--externals` values from the command line argument.

/cc @katywings @krobing

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -476,7 +476,7 @@ function createConfig(options, entry, format, writeMeta) {
 	}
 
 	let globals = external.reduce((globals, name) => {
-		// Use raw/escaped value for CLI-provided RegExp externals:
+		// Use raw value for CLI-provided RegExp externals:
 		if (name instanceof RegExp) name = name.source;
 
 		// valid JS identifiers are usually library globals:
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -465,14 +465,20 @@ function createConfig(options, entry, format, writeMeta) {
...
+			options.external.split(',').map(str => new RegExp(str)),
+		);
 	} else {
 		external = external
 			.concat(peerDeps)
 			.concat(Object.keys(pkg.dependencies || {}));
 	}
 
 	let globals = external.reduce((globals, name) => {
+		// Use raw/escaped value for CLI-provided RegExp externals:
```
katywings, line 479: Do you mean "Use raw/unescaped value"? (Since RegExp.source will not return a "fully" escaped regex https://jsfiddle.net/h6g1s39c/)

developit: Yes, sorry - the comment was confusing/wrong. I think we just want the raw value here?  TBH in the case where an actual regex is provided (`--external .*`) we can't build a useful name anyway - it'll fail the check on L481.

PR #667, opened 2020-06-20
fix: add externals with dash to globals (#643)

katywings:
I should go vacuum-cleaning now, but here is the short description:

## Is
Currently microbundle adds externals to globals - excluding externals with dash (`-`). This results in rollup detecting the missing explicit global and defining the global implicitly with the name camel cased. Also will rollup log a warning like: `No name was provided for external module 'microbundle-test' in output.globals – guessing 'microbundleTest`.

## Should
Since microbundle is adding the externals to globals by default, it should also include those with a dash as mentioned [here](https://github.com/developit/microbundle/issues/643#issuecomment-641071232) - and therefore avoid the `No name was provided` rollup warning. Since rollup handles its dash fallback by adding the external camel cased to globals, microbundle should also do the same.

The only real difference of this change should be, that we don't get the rollup warning `No name was provided for external module` for externals with dash.

## Considerations / Topics for further discussion
- I am not 100% sure how rollups camel casing implementation looks like - literally couldn't find it in rollups src. This means that rollups camelcasing and microbundles is potentially not the same - in this case this could be a breaking change.
- Since this fix is camel casing all externals for the globals, and not only those with a dash, there also is the chance that weird consumer externals like "Blub" will have a different automatic global name after this fix. I don't know any actual dependency that this would hit, but still want to mention it ;).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -482,8 +482,8 @@ function createConfig(options, entry, format, writeMeta) {
 		if (name instanceof RegExp) name = name.source;
 
 		// valid JS identifiers are usually library globals:
-		if (name.match(/^[a-z_$][a-z0-9_$]*$/)) {
-			globals[name] = name;
+		if (name.match(/^[a-z_$][a-z0-9_\-$]*$/)) {
+			globals[name] = camelCase(name);
 		}
 		return globals;
 	}, {});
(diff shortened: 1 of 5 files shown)


developit:
This works great!

PR #668, opened 2020-06-21
Add onStart and onError callbacks, accessible rollup watchers

katywings:
Copied from the commit description ;)

- Adds onStart and onError microbundle js options:
onStart: is called when rollup noticed file changes and is starting with
bundling
onError: called if the rollup watcher is emitting any bundle errors

- Changes microbundle return type to:
```ts
{
  output?: string,
  watchers?: {
    "inputFileName": RollupWatcher
  }
}
```

- Removes obsolete rollup watch FATAL event handling

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -211,8 +211,8 @@ export default async function microbundle(inputOptions) {
 	}
 
 	if (options.watch) {
-		const onBuild = options.onBuild;
-		return new Promise((resolve, reject) => {
+		const { onStart, onBuild, onError } = options;
+		return new Promise(resolve => {
 			stdout(
 				blue(
 					`Watching source, compiling to ${relative(
@@ -221,8 +221,9 @@ export default async function microbundle(inputOptions) {
 					)}:`,
 				),
 			);
-			steps.map(options => {
-				watch(
+
+			const watchers = steps.reduce((acc, options) => {
+				acc[options.inputOptions.input] = watch(
 					Object.assign(
 						{
 							output: options.outputOptions,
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -212,39 +212,37 @@ export default async function microbundle(inputOptions) {
...
+		stdout(
+			blue(
+				`Watching source, compiling to ${relative(
+					cwd,
+					dirname(options.output),
+				)}:`,
+			),
+		);
+
+		return steps.map(options => {
```
wardpeet, line 225: I wonder if people know the order of steps and which watcher is which, maybe it doesn't matter.

katywings: What do you think about reducing the steps to an `entry file name` -> `watcher` object instead of an array? 

Its hard to know which use cases for this might pop up in the future :), i personally only need the array, but a map might be useful for sure - I am open for alternatives if you have an idea :D

katywings: I made some changes: 
- I reverted my change which removed the Promise from microbundle watch. 
- Microbundles promise with enabled watch will now return an object like:

```js
{
  watchers: {
    "/some/absolute/input/filename.js": RollupWatcher
  }
}
```

tools/build-fixture.js
```diff
@@ -60,11 +60,11 @@ export const buildDirectory = async fixtureDir => {
 	process.chdir(resolve(fixturePath));
 
 	const parsedOpts = parseScript(script);
-	let output = '';
-	output = await microbundle({
+	let { output } = await microbundle({
```
katywings, line 63: @developit / @marvinhagemeister This is a breaking change if you consider the microbundle js api "official". On the other hand the js api isn't even mentioned in the README - so idk ^.^. What do you think about it?

marvinhagemeister: For the future it sure sounds reasonable to make the returning value an `object`. I can imagine it being appended with additional information like metadata, sourcemaps, etc.

Agree that we didn't really advertised the JS-API at all. We haven't even done a proper `1.0.0` release yet, so I'm fine with the breaking change here 👍

katywings: Awesome! I updated the commit message and its now officially inofficially a breaking change hehe :+1:

Issue #675, opened 2020-07-04
Type error differences between `tsc` and `microbundle`


tom-sherman:
Reproduction:

- Clone https://github.com/tom-sherman/immurl @ bb2644ecfe1bb4f2098c02df813acf5cf2d5de07
- `npm i`
- `npm run typecheck` - No type errors
- `npm run build` - Type errors from microbundle
- Uncomment methods from bottom of `lib/immurl.ts`
- `npm run typecheck` - Type errors from `tsc`
- `npm run build` - No type errors

Is this a bug or am I missing some configuration? It looks like microbundle is using different lib definitions to typescript here.

marvinhagemeister:
@tom-sherman The repo link leads to a 404 for me. Is this a private repo by chance?

tom-sherman:
@marvinhagemeister Whoops! Should be public now

tom-sherman:
Hmmm, I have found the fix to be to set the "target" in tsconfig.json to "ES6" or above as mentioned [here](https://github.com/Microsoft/TypeScript/issues/23174).

Admitidly this was a mistake on my part, I don't want to target ES5 or below but the error suggests that microbundle isn't respecting my tsconfig?

marvinhagemeister:
Microbundle overwrites both the `module` and `target` compiler options. Those two are the only properties that can't be overwritten.

This is done because we pass the transpiled files back through babel again which does all the downtranspilation for us. Setting it to something lower breaks `import()` syntax. And then we get loads of bugs reports again that `microbundle` would violate it's own no config setup :shrug: 

The default target used by `tsc` when invoked without configuration is `ES5`. There the `URLSearchParam` interface needs less properties to align with the spec.

So there are 2 solutions to this that I can think of:

1) Implement the full interface, instead of coding against the ES5 definition of `URLSearchParams`
2) Specify the lib `.d.ts` files manually (not recommended)

In conclusion 1) is the proper approach that ensures that your code is future proof. We highly recommend going with that :+1:

tom-sherman:
@marvinhagemeister 

> Microbundle overwrites both the module and target compiler options. Those two are the only properties that can't be overwritten.

This is the part I was missing and isn't mentioned in the docs and I don't think many people would say that was obvious. Would you up for reviewing a PR that clarifies this in the README and strongly recommends to match whatever microbundle's target/module is in the tsconfig?

marvinhagemeister:
[...]

PR #676, opened 2020-07-04
Add TypeScript target and module settings explainer

tom-sherman:
This MR adds an explainer to the settings you should set in tsconfig.json lest you run into type errors.

See #675

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ Acts just like `microbundle build`, but watches your source files and rebuilds o
 
 Just point the input to a `.ts` file through either the cli or the `source` key in your `package.json` and you’re done.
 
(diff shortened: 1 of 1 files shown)


katywings:
@tom-sherman I gotta update my project template then which atm just sets tsconfig target: `es5` :).

But if I understand it correctly the tsconfig does not have effects upon the microbundle js output right?

Out of curiosity: which inconsistent type errors did you get without setting target/module to `ESNext`?

tom-sherman:
@katywings Yeah microbundle handles the output so this setting is there just for typechecking via `tsc` and also VSCode.

My type errors were caused by differences in the DOM lib types between ES5 and ESNext target versions, specifically the `URLSearchParams` interface

PR #680, opened 2020-07-11
chore: upgrade all dependencies

wardpeet:
While I was rewriting the microbundle pipeline I needed `preserveEntrySignatures` as an option. This isn't available in rollup@1 so I looked at our dependencies and upgraded them through `npm update`.

In rollup v2, the umd wrapper code has changed to use globalThis as well. This makes the output a little bit bigger.

**BREAKING CHANGES**
<details>
<summary>@rollup/plugin-node-resolve</summary>

v8.0.0
2020-05-20

Breaking Changes
feat: Add default export (#361)
feat: export defaults (#301)

v7.0.0
2020-01-07

Breaking Changes
feat: dedupe by package name (#99)
</details>

<details>
<summary>@rollup/plugin-node-resolve</summary>

v8.0.0
2020-05-20

Breaking Changes
feat: Add default export (#361)
feat: export defaults (#301)

v7.0.0
2020-01-07

Breaking Changes
feat: dedupe by package name (#99)
</details>

<details>
<summary>camelcase</summary>

v6.0.0
2020-04-07

Breaking Changes
Require Node.js 10 (#62) 878bc44
</details>

<details>
<summary>kleur</summary>

v4.0.0
2020-06-17

Breaking Changes
The minimum Node.js runtime increased from 6.x to 10.x since 10.x is the oldest active LTS version.
If you need to continue supporting Node 6.x, either continue using kleur@3.x or ignore the "engines" constraint of kleur@4.x – its CommonJS files will still execute in a Node 6.x environment.
</details>

<details>
<summary>Rollup</summary>

v2.0.0
2020-03-06

Breaking Changes
Rollup now requires at least Node 10 to run, or a sufficiently modern browser (#3346)

The file structure of Rollup's ESM builds has changed:

The main ESM entry point is now at rollup/dist/es/rollup.js instead of rollup/dist/rollup.es.js
The ESM browser build is at rollup/dist/es/rollup.browser.js instead of rollup/dist/rollup.browser.es.js
In general, the ESM builds now follow the same naming scheme as the CJS builds but are located in the rollup/dist/es subfolder instead of rollup/dist (#3391)

The "watch.chokidar" option no longer accepts a boolean value but only an object of parameters that is passed to the bundled Chokidar instance. Chokidar installations by the user will be ignored in favour of the bundled instance (#3331)

Modules that are completely tree-shaken will no longer be listed as part of any chunks in generateBundle

The experimentalOptimizeChunks and chunkGroupingSize options have been removed

acorn plugins can only be used if they accept a passed-in acorn instance instead of importing it themselves. See acornjs/acorn/pull/870#issuecomment-527339830 for what needs to be done to make plugins compatible that do not support this yet (#3391)

Emitted chunks now have the TypeScript type Uint8Array instead of Buffer. A Buffer can still be used, though (#3395)

The TypeScript types no longer use ESTree types for AST nodes but a very generic type that does not contain information specific to certain node types (#3395)

The signature of the writeBundle plugin hook has been changed to match generateBundle: The bundle object is now passed as second parameter instead of first and the first parameter is the output options (#3361)

The following plugin hooks have been removed:

ongenerate: use generateBundle instead
onwrite: use writeBundle instead
transformBundle: use renderChunk instead
transformChunk: use renderChunk instead
You can no longer access this.watcher on the plugin context.

The transform hook can no longer return dependencies.

The treeshake.pureExternalModules option will now show a deprecation warning when used: use treeshake.moduleSideEffects: 'no-external' instead

Using import.meta.ROLLUP_ASSET_URL_<..> and import.meta.ROLLUP_CHUNK_URL_<..> in code will now show warnings: use import.meta.ROLLUP_FILE_URL_<..> instead

The resolveAssetUrl hook will now show a deprecation warning when used: use resolveFileUrl instead

The following plugin context functions will show warnings when used:

this.emitAsset: use this.emitFile
this.emitChunk: use this.emitFile
this.getAssetFileName: use this.getFileName
this.getChunkFileName: use this.getFileName
this.isExternal: use this.resolve
this.resolveId: use this.resolve
Directly adding properties to the bundle object in the generateBundle is deprecated will show a warning (removing properties is allowed, though): Use this.emitFile

Accessing chunk.isAsset on the bundle is deprecated: Use chunk.type === 'asset' instead

The error code for a missing name property when targeting UMD has been changed to MISSING_NAME_OPTION_FOR_IIFE_EXPORT to emphasize this is needed for the IIFE part of UMD (#3393)
</details>

<details>
<summary>rollup-plugin-postcss</summary>

v3.0.0
2020-04-21

Breaking Changes
v3 not support rollup v1.
extract relative path based on bundle root.
</details>

<details>
<summary>rollup-plugin-terser</summary>

v6.0.0
2020-05-23

Breaking Changes
node.js < v10 support is dropped
rollup < v2 support is dropped
sourcemap option is inferred from rollup options, please remove from your config
include/exclude options are dropped in favour of output plugins, see here
</details>

<details>
<summary>rollup-plugin-typescript2</summary>

v0.27.0
2020-03-27

updated dependencies and some type changes

v0.26.0
2020-02-12

#203 Updating object-hash to 2.0.2 with async support.

Updating other dependencies.
</details>

<details>
<summary>tslib</summary>

v2.0.0
2020-05-14

This release changes __exportStar and __importStar to use __createBinding which is incompatible with versions of TypeScript older than 3.9 (i.e. 3.8 and below).

</details>


src/index.js
```diff
@@ -743,6 +742,7 @@ function createConfig(options, entry, format, writeMeta) {
 
 		/** @type {import('rollup').OutputOptions} */
 		outputOptions: {
+			exports: 'named',
```
developit, line 745: Does this break default exports?

test/__snapshots__/index.test.js.snap
```diff
@@ -165,30 +165,30 @@ async-ts
...
+66 B: async-ts.esm.js.br
+186 B: async-ts.umd.js.gz
+143 B: async-ts.umd.js.br"
 `;
 
 exports[`fixtures build async-ts with microbundle 2`] = `7`;
 
 exports[`fixtures build async-ts with microbundle 3`] = `
-"var o=function(){function o(){}return o.prototype.foo=function(){return Promise.resolve()},o}();export{o as MyClass};
+"class e{foo(){return Promise.resolve()}}export{e as MyClass};
```
developit, line 179: Seems like something is causing us to output modern code in non-modern bundles. Not sure which dep but my guess is `preset-env` or `typescript`?

test/__snapshots__/index.test.js.snap
```diff
@@ -165,30 +165,30 @@ async-ts
...
 //# sourceMappingURL=async-ts.esm.js.map
 "
 `;
 
 exports[`fixtures build async-ts with microbundle 4`] = `
-"exports.MyClass=function(){function o(){}return o.prototype.foo=function(){return Promise.resolve()},o}();
+"exports.MyClass=class{foo(){return Promise.resolve()}};
```
developit, line 185: shouldn't be using class here.

test/__snapshots__/index.test.js.snap
```diff
@@ -220,30 +220,30 @@ basic
...
+131 B: basic-lib.esm.js.br
+270 B: basic-lib.umd.js.gz
+214 B: basic-lib.umd.js.br"
 `;
 
 exports[`fixtures build basic with microbundle 2`] = `6`;
 
 exports[`fixtures build basic with microbundle 3`] = `
-"var r=function(){try{var r=arguments;return Promise.resolve([].slice.call(r).reduce(function(r,e){return r+e},0))}catch(r){return Promise.reject(r)}};export default function(){try{var e=arguments,t=[].slice.call(e);return Promise.resolve(r.apply(void 0,t)).then(function(e){return Promise.resolve(r.apply(void 0,t)).then(function(r){return[e,r]})})}catch(r){return Promise.reject(r)}}
+"const e=function(){try{return Promise.resolve([].slice.call(arguments).reduce((e,r)=>e+r,0))}catch(e){return Promise.reject(e)}};export default function(){try{var r=[].slice.call(arguments);return Promise.resolve(e(...r)).then(function(t){return Promise.resolve(e(...r)).then(function(e){return[t,e]})})}catch(e){return Promise.reject(e)}}
```
developit, line 234: const and arrow function in a non-modern bundle

developit:
My guess is that we are using some babel deps that still use the "whitelist" term, which was changed to "allowlist". The [changelog](https://github.com/babel/babel/releases) makes it seem like this was an internal-only change, but if we have a mix of versions it could cause issues.

wardpeet:
Yeah, I'm going to update one by one to figure it out :)

src/index.js
wardpeet: Unsure yet, I was just removing warnings 😛

wardpeet:
One thing I noticed is that the css chunk keeps changing, it's not deterministic and it seems like moduleIds from rollup is the culprit.

developit:
@wardpeet I wonder if its the CSS Modules identifier generation we're doing custom?

wardpeet:
I've created a PR in postcss-plugin for rollup to fix this issue
https://github.com/egoist/rollup-plugin-postcss/pull/295/

JChiaramonte7:
can we roll #711 into this?

rblalock:
This going to be merged soon? There's a high vulnerability in the rollup-plugin-terser that is fixed in a later version.

wardpeet:
Upstream changes have landed, so it should land relatively soon. I need to find the time to make sure it's 100% working.

@rblalock you can checkout https://github.com/developit/microbundle/issues/695. You shouldn't worry to much about the vulnerability as microbundle is only used during build time.

src/index.js
```diff
@@ -80,24 +79,13 @@ export default async function microbundle(inputOptions) {
...
-					options,
-					options.entries[i],
-					formats[j],
-					i === 0 && j === 0,
-				),
-			);
-		}
+	for (let j = 0; j < formats.length; j++) {
+		steps.push(createConfig(options, options.entries, formats[j], j === 0));
 	}
```
wardpeet, line 89: All entries are moved into 1 rollup pass per format so code-splitting can happen.

src/index.js
```diff
@@ -245,64 +235,72 @@ function replaceName(filename, name) {
...
+			pkg['umd:main'] || pkg.unpkg || 'x.umd.js',
+			mainNoExtension,
+		);
+
+		const file = mainsByFormat[format] || mainsByFormat.cjs;
+		inputs[basename(file, extname(file))] = entry;
+	});
 
-	return mainsByFormat[format] || mainsByFormat.cjs;
+	return inputs;
```
wardpeet, line 290: This needs some cleanup but this should be done in a separate PR where we decouple more.

wardpeet (commented):
Everything works except with merging inputs into 1 pass. Multiple inputs are not allowed anymore when using UMD or iife. It will throw an exception. I cannot fix that part. Users will have to run microbundle twice.

This is a breaking change. I couldn't find a way to disable code splitting which leads to this issue.

PR #685, opened 2020-07-16
upgrade postcss to fix automodules (fixes #653)

### LINKED ISSUE #653: All CSS-files are treated as CSS Modules
Regardless `--css-modules` flag, it is impossible to use CSS and CSS Modules in the same project. It seems that `null` option works incorrectly and Rollup treats all CSS files as modules.

To reproduce, you could try to build this simple example:

```jsx
import React from 'react';

import './index.css'
import styles from './index.module.css';

const App = () => (
  <div>
    <div className={styles.moduleClass}>CSS module class</div>
    <div className="local-no-module-class">No-module class</div>
  </div>
);

export default App;
```

Building with `microbundle ./index.jsx` makes the following `index.css` file:
```css
._1fWqO{color:#00f}
._3umtn{color:red}
```

However, the expected result is: 

```css
._1fWqO{color:#00f}
.local-no-module-class{color:red}
```

developit:


katywings:
Looks like its failing for this test: https://github.com/developit/microbundle/tree/master/test/fixtures/css-modules--string

Not completely sure yet what is going on with the differing snapshots :thinking:

developit:
Weird - JS sizes went up too.

developit:
Turns out our snapshots were just incorrect.

wardpeet:
@developit Sadly this cannot be done because rollup v1 is not supported anymore
https://github.com/egoist/rollup-plugin-postcss/releases/tag/v3.0.0

We're currently blocked by https://github.com/egoist/rollup-plugin-postcss/pull/295 to update microbundle to rollup v2. I could update https://github.com/developit/microbundle/pull/680 to use a git package instead or publish the patch under @wardpeet npm package if that helps.

wardpeet:
Going to close this because https://github.com/developit/microbundle/pull/680 will fix this once new version of postcss is being released.

PR #692, opened 2020-08-08
Document @ for the --define option

gregorym:
I had to dig through the code in order to figure out how to replace an expression.
I was initially trying to do `microbundle --define global=window` which should actually be `microbundle --define @global=window`


wardpeet:
@gregorym Thank you for opening this PR. Could we instead add a small section to https://github.com/developit/microbundle#-usage--configuration- that explains how to do definitions with examples? 

https://github.com/developit/microbundle/blob/58813871f490d0ec32bcb7d734cbc89fe2482eba/src/lib/option-normalization.js#L5-L24

developit (approved):

PR #700, opened 2020-08-16
fix: disable node-builtin warnings

wardpeet:
Disables warnings for node's builtin modules like stream, child_process when target is node. Rollup-node-resolve marks these modules as external but still shows a warning.

```
  console.warn node_modules/rollup/dist/shared/node-entry.js:13595  
    'child_process' is imported by src\index.js, but could not be resolved – treating it as an external dependency
```


katywings:
@wardpeet Awesome work! :)

Issue #694, opened 2020-08-10
Optional Chaining Breaks Without No Compress Flag


pkrawc:
Running `microbundle --jsx React.createElement` will break on the following component unless `--no-compress` is passed

```jsx
import * as React from "react"
import { createPortal } from "react-dom"

export default function Portal({ children, type = "avail-portal" }) {
  const portalNode = React.useRef(null)
  const mountNode = React.useRef(null)
  const [s, update] = React.useState({})
  React.useLayoutEffect(() => {
    if (typeof window === "undefined" || !mountNode.current) return
    const owner = mountNode?.current?.ownerDocument
    portalNode.current = owner.createElement(type)
    owner.body.appendChild(portalNode.current)
    update({})
    return () => {
      if (portalNode?.current?.ownerDocument) {
        portalNode?.current?.ownerDocument?.body?.removeChild(
          portalNode.current,
        )
      }
    }
  }, [mountNode])
  return portalNode.current ? (
    createPortal(children, portalNode.current)
  ) : (
    <aside ref={mountNode} />
  )
}
```

developit:
I believe this is because we're using `@babel/preset-modules` for the modern builds. We need to switch to `@babel/preset-env` now that it has the `bugfixes:true` option.

pkrawc:
Thanks @developit for the workaround for now. Will microbundle consume `babel.config.js` the same way as a `.babelrc`?

pkrawc:
I assume yes but I've only ever seen `.babelrc` mentioned in the documentation.

wardpeet:
@pkrawc we should also support babel.config.js if not mistaken.

Issue #522, opened 2019-11-16
TypeScript 3.7.x: Optional chaining and nullish coalescing is broken


smithki:
The latest version of TypeScript (`3.7.x`) is broken if using optional chaining and nullish coalescing. It fails with the following error:

```sh
SyntaxError: /path/to/file.ts: Support for the experimental syntax 'optionalChaining' isn't currently enabled

...

Add @babel/plugin-proposal-optional-chaining (https://git.io/vb4Sk) to the 'plugins' section of your Babel config to enable transformation.
```

The simple fix (I think) is to update the TypeScript dependency in `microbundle`. _However..._ my preference would be for `microbundle` to detect if I have another version of TypeScript already registered in my `package.json` and _use that version_ as the default (falling back to the included dependency if required).

smithki:
I spoke too soon. After digging a little deeper, it seems the issue is actually related to TypeScript emitting optional chaining and nullish coalescing without transpilation with `esnext`... Until the babel plugin above is added to `microbundle`, is there a local workaround I can use? It doesn't appear that I can extend the babel config...

I updated the title to reflect this new information.

tunnckoCore:
> to detect if I have another version of TypeScript already registered in my

Microbundle doesn't do anything special, it uses the typescript plugin which is responsible for handling that.

> is there a local workaround I can use?

I guess switch the `esnext`? I'm not sure.

ForsakenHarmony:
You can add a custom babelrc (the first babel plugin is for constants)

smithki:
I figured out from reading the source that a a custom babelrc is supported in the next version. I think this can be closed now... though I wonder, is it out of scope for `microbundle` to allow TypeScript to handle all transpiration? My overall preference would be for `microbundle` to respect the value of `target` in my `tsconfig.json`... is there a reason why that is not the case?

ConsoleTVs:
I came to the same conclusion. Shame, it was my first shot on this lib. It's a pretty it does not merge babel settings or at least, allow you to customize it or append plugins. Anyway, this should be a top priority because as for today, the statement: **Typescript supported**, it's not true. Will keep an eye here for the fix. And if the fix is already there in the new version, why not publish it? @developit

ForsakenHarmony:
@smithki I'd assume you already are on the next version, because we didn't have babel in 0.11.0

It should work if you add a `.babelrc` to your project with this content

```json
{
	"plugins": [
		[
[...]

PR #702, opened 2020-08-16
fix: use @babel/preset-env with bugfixes for modern builds

wardpeet:
This PR removes the usage of `@babel/preset-modules` and use `@babel/preset-env` with the bugfixes flag when bundling for modern browsers.

This enables "Optional chaining" and "nullish coalescing" by default without a babelrc override.

Fixes https://github.com/developit/microbundle/issues/522
Fixes https://github.com/developit/microbundle/issues/694


src/lib/babel-custom.js
```diff
@@ -46,7 +46,8 @@ const createConfigItems = (babel, type, items) => {
 	});
 };
 
-const presetEnvRegex = RegExp(/@babel\/(preset-)?env/);
+const environmentPreset = '@babel/preset-env';
```
developit, line 49: This no longer finds `@babel/env`, which is an alias of `@babel/preset-env`

src/lib/babel-custom.js
```diff
@@ -141,6 +138,7 @@ export default () => {
 									},
 									preset.options,
 									{
+										bugfixes: customOptions.modern,
```
developit, line 141: This can just always be set to `true`, it has no effect on cases where the output is below the affected syntax level.

src/lib/babel-custom.js
wardpeet: Oh I searched for it and couldn't find any usage of it. i'll create a follow up PR

src/lib/babel-custom.js
wardpeet: Sweet

src/lib/babel-custom.js
developit: Heh - yeah we don't have a test for it, just it's possible to get that value when we read in user config.

jdpigeon:
🎉  thank you!

probablyup:
Can this be released? Just ran into https://github.com/developit/microbundle/issues/522 and wasted some time...

PR #706, opened 2020-08-19
Initial Eik adaptations

stipsan:
The main additions are the eik rollup plugin, and changes to the replaceName function to allow packages to set custom filenames to the output in dist, instead of being forced to stay within the microbundle naming conventions with `[pkg.name].[format].js`.


developit:
@stipsan FWIW I'd be open to supporting Import Maps in Microbundle.

stipsan:
@developit that's awesome! We're starting to test import maps and esm with shared libraries on the cdn in production. We're expecting to do some iterations on this before we land on a setup that works well. 
The intentions of our fork is to be able to try out different approaches quickly, but the endgame is we'd like to collab on landing changes upstream so that the fork is no longer necessary 🙂 
/cc @trygve-lie @digitalsadhu

PR #714, opened 2020-09-01
fix: bring back @babel/env preset match

wardpeet:
@developit added @babel/env check again to microbundle. Npm doesn't have the package https://www.npmjs.com/package/@babel/env. It's the reason why I removed it so unsure what alias is meant in https://github.com/developit/microbundle/pull/702#discussion_r479397230. I'll change the comment in this PR to be more descriptive when I get more info.

Got removed in https://github.com/developit/microbundle/pull/702#discussion_r479397230

Note: this does not need a changeset

### DIFF
--- a/src/lib/babel-custom.js
+++ b/src/lib/babel-custom.js
@@ -47,7 +47,7 @@ const createConfigItems = (babel, type, items) => {
 };
 
 const environmentPreset = '@babel/preset-env';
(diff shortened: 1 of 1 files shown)


wardpeet:
@developit where can I find more info on the @babel/env alias part?

developit:
@wardpeet https://babeljs.io/docs/en/presets#preset-shorthand

PR #727, opened 2020-09-14
docs: Fix simple typo, realy -> really

timgates42:
There is a small typo in test/index.test.js.

Should read `really` rather than `realy`.


marvinhagemeister (approved):
Good catch! :raised_hands:

PR #731, opened 2020-09-23
Add jsxImportSource flag for new JSX runtime

vaneenige:
This is an update to support the new JSX runtime in the preact@10.5.0 release! If you don't provide the flag it'll work as normal with the default or provided JSX pragma. If it it provided and for example "preact" it'll work in the new way. 🎉

More info here: https://github.com/preactjs/preact/releases/tag/10.5.0

### DIFF
--- a/.changeset/quick-lies-buy.md
+++ b/.changeset/quick-lies-buy.md
@@ -1,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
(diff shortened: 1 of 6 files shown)


ForsakenHarmony (approved):

marvinhagemeister (approved):
Sweet! Thank you so much for the PR :raised_hands:

PR #739, opened 2020-10-14
[WIP] Add Vue support

### LINKED ISSUE #96: Add Vue.js support
I've recently built a [Vue.js component](https://github.com/alexjoverm/v-lazy-image) for lazy loading images and wanted to use microbundle for its setup, but couldn't make it work.

What's required to have Vue.js support?

wladi0097:
Closes #96 

There are still issues with the compatibility with Typescript. But the regular Vue compilation already works as expected.
The main idea is that we can use [rollup-plugin-vue](https://rollup-plugin-vue.vuejs.org/) plugin to compile [class-components](https://class-component.vuejs.org/).

What do you think ?


developit:
Hi @wladi0097! Thanks for the PR. I worry that supporting custom formats moves too far away from Microbundle's goal of being a simple optimized bundler for JS modules. Because we can't provide a solution for CSS that avoids bundling it into JavaScript, the premise of Single File Components feels a bit misleading - we can't really generate output that would be in line with Microbundle's goal of producing small bundles.

There's ongoing discussion about whether we should allow rollup plugins, perhaps this is another case in favor of doing so.

wladi0097:
Understandable, closing PR.

Issue #711, opened 2020-08-28
Unknown compiler option 'jsxFragmentFactory'


TechQuery:
I guess MicroBundle hasn't support TypeScript 4.x, because I got an error:
> rpt2: config error TS5023: Unknown compiler option 'jsxFragmentFactory'.

wardpeet:
Thanks will look into it 👍 

Could you create a small comment on how to reproduce this? Or do you have a repo?

TechQuery:
Just add an option `"jsxFragmentFactory": "Fragment"` in your `tsconfig.json`, and then run `microbundle` again, you'll see the error.

developit:
We should check TS version and use the appropriate config - it would be a shame to have Microbundle only work with TS 4.

whitetrefoil:
Not only tsconfig.json options but also code syntax will have the same ts version problem.

In my case:
```ts
try {
  something()
} catch (e: unknown) {
         ^ (rpt2 plugin) Error: semantic error TS1196: Catch clause variable cannot have a type annotation.
  something()
}
```

Where here `(e: unknown)` is [new syntax in TS 4.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#unknown-on-catch-clause-bindings), it's recommanded (safer) and enforced by eslint.

Maybe microbundle can try locally installed ts first before using rpt2's default one.  Or at least some way to set the version manually.

developit:
Thanks for the idea and PR @whitetrefoil - good thinking.

developit:
Fixed in #738.

PR #741, opened 2020-10-16
Use user's typescript first, fallback to bundled

whitetrefoil:
This PR is aimed at fixing #711.

Microbundle will load "typescript" from `path.dirname(process.argv[1])`.

If no `process.argv[1]`, or failed to resolve "typescript" from there, fallback to the bundled "typescript".

P.S.
Using `resolve-from@v4.0.0` instead of the latest `v5.0.0` because `v4.0.0` has already been in `package-lock.json`.

### DIFF
--- a/.changeset/silver-guests-brake.md
+++ b/.changeset/silver-guests-brake.md
@@ -1,0 +1,5 @@
+---
+'microbundle': minor
+---
+
+Use user's typescript first, fallback to bundled
--- a/src/index.js
+++ b/src/index.js
@@ -487,8 +487,10 @@ function createConfig(options, entry, format, writeMeta) {
 					},
 					useTypescript &&
 						typescript({
-							typescript: require(resolveFrom(process.cwd(), 'typescript') ||
-								'typescript'),
+							typescript: require(resolveFrom(
+								dirname(process.argv[1]),
+								'typescript',
+							) || 'typescript'),
 							cacheRoot: `./node_modules/.cache/.rts2_cache_${format}`,
 							useTsconfigDeclarationDir: true,
 							tsconfigDefaults: {
--- a/src/index.js
+++ b/src/index.js
@@ -403,6 +403,18 @@ function createConfig(options, entry, format, writeMeta) {
 	const outputDir = dirname(absMain);
 	const outputEntryFileName = basename(absMain);
 
+	const argv1 = process.argv[1];
+	let typescriptLibPath;
+	if (argv1 != null) {
+		// If argv1 exists, try to require 'typescript' from there.
+		typescriptLibPath = resolveFrom(dirname(argv1), 'typescript');
+	}
+	if (typescriptLibPath == null) {
+		// If argv1 doesn't exist or no 'typescript' found there,
+		// fallback to bundled 'typescript'.
+		typescriptLibPath = 'typescript';
+	}
+
 	let config = {
 		/** @type {import('rollup').InputOptions} */
 		inputOptions: {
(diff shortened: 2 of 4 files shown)


wardpeet:
@whitetrefoil, could you explain a little what the origin is of this feature? I bet you want to use typescript 4.0 and we don't support it yet?

whitetrefoil:
> @whitetrefoil, could you explain a little what the origin is of this feature? I bet you want to use typescript 4.0 and we don't support it yet?

Yes, you're correct. See #711 where has 2 examples there: `jsxFragmentFactory` and `catch (e: unknown)`.

BTW, IMO use the ts version of the project to be bundled should make more sense than using a fixed version.

src/index.js
```diff
@@ -402,6 +403,18 @@ function createConfig(options, entry, format, writeMeta) {
 	const outputDir = dirname(absMain);
 	const outputEntryFileName = basename(absMain);
 
+	const argv1 = process.argv[1];
+	let typescriptLibPath;
+	if (argv1 != null) {
+		// If argv1 exists, try to require 'typescript' from there.
+		typescriptLibPath = resolveFrom(dirname(argv1), 'typescript');
```
developit, line 410: I think this could be changed to use the cwd rather than assuming microbundle is run from the cwd. It also needs to use `.silent()`, otherwise this throws when there's no local typescript.

```suggestion
		typescriptLibPath = resolveFrom.silent(options.cwd, 'typescript');
```

developit (changes requested):
Looks good! I think we can simplify this down to basically the following:

`require(resolveFrom.silent(options.cwd, 'typescript') || 'typescript'))`

... but otherwise good to go.

src/index.js
whitetrefoil: You're right. There's no reason to prevent project's "typescript" to be required by global microbundle.  Applied your simplified solution.

src/index.js
```diff
@@ -9,6 +9,7 @@ import autoprefixer from 'autoprefixer';
 import cssnano from 'cssnano';
 import { rollup, watch } from 'rollup';
 import builtinModules from 'builtin-modules';
+import resolveFrom from 'resolve-from';
```
ForsakenHarmony, line 12: why do we need this instead of just `require.resolve` and will this break with something like yarn pnp / pnpm?

whitetrefoil: Not much different in theory. "resolve-from" needs node v8.0 while `paths` options of `require.resolve` requires node v8.9.  And "resolve-from" support `.silent` which will lead to a bit nicer code in this case.

This PR works with yarn, I'm using yarn actually.  But I have no idea about pnpm.

BTW there's [an issue in "require-resolve"](https://github.com/sindresorhus/resolve-from/issues/8) which complained that `paths` of `require.resolve` doesn't work. But I never dug deeper about that.

ForsakenHarmony: Node 8 is EOL so that doesn't matter

You're using yarn 2? Then it's fine to merge

ForsakenHarmony (approved):

.changeset/silver-guests-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'microbundle': minor
```
ForsakenHarmony, line 2: Do you consider this a breaking change?

whitetrefoil: Used to considered it to be a fix + feature instead of break, and it can fallback to previous behavior.

After thought more carefully... Maybe you're right, in theroy if user previously used different TS version during developing & bundling, it will lead to different result.

But I'm still hesitating. Logically if a user previously:
* used same version in dev & bundling - no change
* verA in dev + verB in bundling, both worked - use verA in both stages should also work
* verA in dev + verB in bundling, OK in dev but failed to bundle - this PR means a fix i.o. a break to this user
* verA in dev + verB in bundling, OK in bundling but failed in dev - this PR is a break, but I'm not sure if this just happen...

Anyway, major is always a safer choice... Shall I update the PR?

developit (approved):

PR #750, opened 2020-11-09
Update examples to reference preact/compat instead of preact

camsloanftc:
aliasing react=preact doesn't work because the main preact package actually doesn't contain any default exports. So the more likely example would be to reference preact/compat.


marvinhagemeister (approved):
Great catch, thanks a bunch for the PR 🙌

Issue #679, opened 2020-07-08
How can I include SVGs in a build?


bkiac:
I import SVG files in my React project like this:
```js
import { ReactComponent as Icon } from "../../assets/bee-box-logo.svg";
```
but the build fails with this error message:
```bash
$ microbundle --jsx React.createElement
Error: Unexpected token (Note that you need plugins to import files that are not JavaScript)

at /home/bkiac/Repos/space-apps/honeycomb/src/assets/bee-box-logo.svg:1:0

1: <svg width="67" height="48" viewBox="0 0 67 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```

How can I include SVGs in my microbundle build? Can I extend the microbundle config and add plugins to the underlying rollup config?

marvinhagemeister:
Nope, for that we recommend going with `rollup` directly and writing a config for your needs. Another option is to paste the svg into your JavaScript files as a React component. That is usually beneficial if you need to colorize icons as you have now the additional option of using `currentColor`.

katywings:
@bkiac An other alternative could be to pregenerate your svgs into react components with https://github.com/gregberge/svgr

ForsakenHarmony:
I do something similar with https://github.com/ForsakenHarmony/preact-feather, but it's a bit more manual than i.e. svgr and I only use microbundle there for cjs and umd

olance:
@marvinhagemeister I have the same need as @bkiac and so am wondering: any "easy" path from microbundle to rollup in that case? 
Any way to get a ready-to-use rollup config from microbundle for instance?

marvinhagemeister:
@olance The rollup config from microbundle is declared here: https://github.com/developit/microbundle/blob/a6eb6e75b0bea09ddcb127689cde77b4e75f86f2/src/index.js#L405

olance:
Thanks! Indeed I've taken the time to actually search and found that quite easily, sorry for asking before searching 🙈

PR #755, opened 2020-11-21
feat(config): allow override of rollup & plugins config by project config

olance:
I love the simplicity of microbundle, but I need to do a bit more than what the default configuration it builds up provides.

More specifically, I want to be able to import SVG images.
Issue #679 says "for that we recommend going with rollup directly and writing a config for your needs.".

Given everything microbundle does for me, having to let it go and configure rollup myself just because of some SVG files makes me quite sad, so I pulled up this PoC of how microbundle could allow some customization of its build config.

@marvinhagemeister as you were the one making that statement in the SVG issue, maybe you'd be able to review/comment this?

It's quite "quick and dirty" for the moment, but I first wanted to know whether that could be a direction microbundle could take?

As a concrete example, that PR allows me to add a `microbundle.config.js` file into my project and support importing SVG files by importing the `@rollup/plugin-image` plugin:

```js
const image = require('@rollup/plugin-image')

module.exports = {
  plugins: {
    typescript: function (config) {
      // Since imported *.svg will be transformed to ES modules by the image plugin, tell
      // TypeScript to process them
      return { ...config, include: ['*.ts+(|x)', '**/*.ts+(|x)', '*.svg', '**/*.svg'] }
    },
  },

  config: function (config) {
    // Add image plugin right after json plugin
    const jsonIdx = config.inputOptions.plugins.findIndex((p) => p.name === 'json')
    config.inputOptions.plugins.splice(jsonIdx + 1, 0, image())
    return config
  },
}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -29,6 +29,7 @@ import {
 } from './lib/option-normalization';
 import { getConfigFromPkgJson, getName } from './lib/package-info';
 import { shouldCssModules, cssModulesConfig } from './lib/css-modules';
+import { getConfigOverride } from './lib/config-override';
 
 // Extensions to use when resolving modules
 const EXTENSIONS = ['.ts', '.tsx', '.js', '.jsx', '.es6', '.es', '.mjs'];
@@ -92,7 +93,7 @@ export default async function microbundle(inputOptions) {
 	for (let i = 0; i < options.entries.length; i++) {
 		for (let j = 0; j < formats.length; j++) {
 			steps.push(
-				createConfig(
+				await createConfig(
 					options,
 					options.entries[i],
 					formats[j],
@@ -291,8 +292,9 @@ function getMain({ options, entry, format }) {
 // shebang cache map because the transform only gets run once
 const shebang = {};
 
-function createConfig(options, entry, format, writeMeta) {
+async function createConfig(options, entry, format, writeMeta) {
 	let { pkg } = options;
+	const context = { options, entry, format, writeMeta };
 
 	/** @type {(string|RegExp)[]} */
 	let external = ['dns', 'fs', 'path', 'url'].concat(
@@ -402,6 +404,8 @@ function createConfig(options, entry, format, writeMeta) {
 	const outputDir = dirname(absMain);
 	const outputEntryFileName = basename(absMain);
 
+	const configOverride = await getConfigOverride(context);
+
 	let config = {
 		/** @type {import('rollup').InputOptions} */
 		inputOptions: {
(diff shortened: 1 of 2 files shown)


olance:
Any feedback on this? @developit maybe?

developit:
Hi @olance! I haven't had a chance to review the code here yet, but I am struggling with the philosophical/directional impact.

Microbundle was created to serve a very specific need: bundling libraries before they are published to npm. I know lots of folks are using it for much more than that, but I don't actually think it serves other use-cases very well. It doesn't provide hot module replacement, bakes in relatively opinionated default output configurations, includes default plugins that can't be un-defaulted, compiles code in an aggressively loose configuration, etc. These things all make sense for bundling libraries, but they're not really great defaults for apps or websites, and they would make configuration destructive (as you pointed out with the plugin ordering snippet, plugins have to be aware of other plugins).

On the flip side, we've recently released [WMR](https://github.com/preactjs/wmr). It's a standalone application development tool that uses Rollup to bundle for production and supports Rollup plugins (during development too, despite not bundling your source code). Since WMR is maintained by roughly the same group of people who maintain Microbundle, it feels like we might be better off transitioning folks currently using Microbundle as an app bundler over to WMR.

Ok, coming back to libraries (perhaps you're wanting to bundle these SVG's into a library?). One of the things I'm stuck on with this PR is that it seems like the example could be addressed in a generalized way without introducing configuration _(more on that in a minute)_. Microbundle derives immense benefit from not offering configuration, because it means we don't leak the details of dependencies like Rollup to users. This allows controlling versions and compatibility, which is what makes it possible to upgrade microbundle across many projects without breaking configurations. One example where this is beneficial is the upgrade to Rollup 2 - we're held off on upgrading because the bundles produced were larger than before. If we had leaked Rollup or dependencies by allowing configuration, we wouldn't have caught those changes and folks likely would have ended up with larger bundles just from updating devDependencies.

Looking at your specific SVG use-case, what if Microbundle were to offer a `text:` import prefix that returned the text content of the filepath it were applied to?

```js
import svg from 'text:./icon.svg';
     // ^ "<svg ....</svg>"

export const Icon = () => <span className="icon" dangerouslySetInnerHTML={{ __html: svg }} />;
```

keul:
@developit whatever will be the final implementation what is missing is quite clear: bundling a resource with a package (an image, or CSS background image, or a font) using microbundle is very hard (or not possible).

marvinhagemeister:
@keul To give some context: That scenario is hard because microbundle was never build to bundle resources. It was born out of our main Preact repo where folks always asked us if we could turn our build pipeline into something that could be re-used for other libraries.

developit:
Indeed. One thing I forgot to mention in my reply that might help clarify: the original reason Microbundle did not include a way to bundle non-JS resources is because there's no standardized mechanism to do so. The closest thing we have to a standard for importing non-JS resources is a few bundlers using a similar import prefix technique.

Then there's the issue of bloat: it's certainly most convenient to bundle images or fonts into JS when publishing to npm, but the performance impact of doing so is surprisingly dire. Images and fonts are binary data that must be Base64-encoded to embed in JS as strings. Embedding a 300kb font in an npm module will increase everyone's bundle size by 1mb, because Base64 incurs approximately 2x overhead. Worse still, it's not just the file size that becomes a problem - 1mb of image data in a JavaScript bundle incurs parsing and execution cost, because there is no way for a JS engine to know a given string is 1mb of external data.

Since Microbundle's primary goal is to help folks bundle modules in a way that keeps them small and fast, making it easy to explode bundle sizes sortof goes against the project's purpose.

That said, I totally agree there's a need for _something_ here. The core of the problem is that there is no accepted way to refer to resources like images or fonts from within an npm module, which is what leads folks to go the embedding route. Maybe we can change that? In a pure ES Modules world, the follow code works:

```js
const img = document.createElement('img');
img.src = new URL('./my-image.svg', import.meta.url);
document.body.appendChild(img);
```

The SVG is kept out of JavaScript, and bundlers can transform or statically evaluate the `URL()` construction. I believe Webpack is now doing this by default in version 5, which gives me hope that this could work in practise.

keul:
Thanks for your point on this.

I perfectly understand the issue, my general concern was exactly about this: no shared way of doing this, although usecases where external a JavaScript library should provide also a resource (or a CSS) is quite common today.

developit:
Yeah, CSS really seems like the #1 pain point here. As part of thinking about this, I've opened preactjs/wmr#260 to land support for bundling the above technique there (if it's not already supported, which is might be? haha).

PR #757, opened 2020-11-24
feat(web-workers): Add rollup-plugin-web-worker-loader

schlusslicht:
**What this pull request does**
This pull request aims at providing a convenient way to create inlined JavaScript source blobs which contain all dependencies to run within a WebWorker execution context.

**What this pull request contains**
- implementation ([here](https://github.com/schlusslicht/microbundle/blob/feat/worker-loader/src/index.js#L588-L592))
- worker-loader cli option ([here](https://github.com/schlusslicht/microbundle/blob/feat/worker-loader/src/prog.js#L59))
- TypeScript-enabled test case ([here](https://github.com/schlusslicht/microbundle/tree/feat/worker-loader/test/fixtures/worker-loader))
- small documentation change/example ([here](https://github.com/schlusslicht/microbundle/tree/feat/worker-loader#building-web-workers-as-inline-blobs))
- repository to test the implemented feature "in the wild" ([here](https://github.com/schlusslicht/rollup-typescript-webworkers/tree/demo/microbundle))

**Further notes**
I've tried to adhere to the standards of the codebase and the resulting changes can successfully build the [rollup-typescript-webworkers](https://github.com/darionco/rollup-typescript-webworkers) repository, which is an example how to use the [rollup-plugin-web-worker-loader](https://github.com/darionco/rollup-plugin-web-worker-loader). Feel free to point me in the right direction if You consider to merge this, but need some changes to be done! :)

**Issues/drawbacks**
In order to successfully use the rollup-plugin-web-worker-loader, I had to disable the Rollup build cache when running microbundle with the plugin enabled (see [here](https://github.com/schlusslicht/microbundle/blob/feat/worker-loader/src/index.js#L401)). This seems to be related to an incompatibility between this plugin and something else. I've tried to investigate further, but dropped the lead after spending to much time digging through the Rollup internals. As for now, the behavior of microbundle/Rollup build cache does not change, when building without this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,29 @@ By default Microbundle outputs multiple bundles, one bundle per format. A single
 microbundle -i lib/main.js -o dist/bundle.js --no-pkg-main -f umd
 ```
 
+### Building web workers as inline blobs
+
+Through the [rollup-plugin-web-worker-loader](https://github.com/darionco/rollup-plugin-web-worker-loader), microbundle can be instructed to create inlined JavaScript source blobs which contain all dependencies to run within a WebWorker execution context. Consider the following minimal example:
+
+```js
+// main.js
+import MyWorker from 'web-worker:./my-worker';
+const myWorker = new MyWorker();
+myWorker.postMessage(3.14159);
+
+// my-worker.js
+import { longRunningJob } from 'heavy-calculations'; // Imports will be inlined
+self.onmessage = message => self.postMessage(longRunningJob(message.data));
+```
+
+And build it like this:
+
+```bash
+microbundle --worker-loader
+```
+
+**Note** For usage in TypeScript projects, have a look at the [rollup-typescript-webworkers](https://github.com/darionco/rollup-typescript-webworkers) example by the rollup-plugin-web-worker-loader maintainers.
+
 ### Mangling Properties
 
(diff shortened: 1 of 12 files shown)


developit:
Thanks for the comprehensive PR. I'm a fan of using workers and making worker usage easier. That said, I'm not sure a prefix is the right direction for Microbundle to take. It's specific to `Worker`, which means we'd need to add plugins if we ever wanted to provide the identical functionality for worklets, SharedWorker, `worker_threads`, frames, etc. 

A syntax detection-based approach like that of [rollup-plugin-off-main-thread](https://github.com/surma/rollup-plugin-off-main-thread#auto-bundling) or [rollup-plugin-worker](https://github.com/johanholmerin/rollup-plugin-worker) would get around this concern. Ultimately, I think it might make more sense to land support for the general-purpose `import workerURL from 'bundle:./worker.js'` import prefix, since that can be easily used to instantiate workers and worklets.

Thoughts?

schlusslicht:
Thank You for Your review and feedback, Jason! I'm currently engaged in other projects, but will (re)evaluate this PR/funcionality some time in the future. Unless someone is faster :smiley:. Would like to give You an ETA, but my timeline is too messy currently.

developit:
This landed in #815 / #867.

PR #760, opened 2020-11-28
Upgrade TypeScript version to 4.1.2

oriSomething:
I've listed as a `minor` instead of `patch` (Because TS isn't semver). Event though in rare cases it might be a `major` change.
I've tested it on a personal project that uses TS 4.1.

### DIFF
--- a/.changeset/loud-days-notice.md
+++ b/.changeset/loud-days-notice.md
@@ -1,0 +1,5 @@
+---
+'microbundle': minor
+---
+
+Upgrade TypeScript version to 4.1.2
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "microbundle",
-  "version": "0.12.3",
+  "version": "0.12.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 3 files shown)


vladmiller:
Currently, with yarn I am forcing ^4.1 version using resolution in package.json

```
"resolutions": {
  "microbundle/typescript": "^4.1"
},
```

however, it throws an error when using `react-jsx` option.
```
rpt2: options error TS5089: Option 'jsxFactory' cannot be specified when option 'jsx' is 'react-jsx'.
```

I think more investigation required here.

rschristian:
> 
> Here is where `jsxFactory` set by minibundle: https://github.com/developit/microbundle/blob/master/src/index.js#L498-L503
> 
> The workaround is to set jsxFactory inside of tsconfig.json; however, maybe microbundle should try to detect settings from tsconfig.json and don't try to set jsxFactory if jsx is set to react-jsx?

Not quite. 

Microbundle passes a set of defaults to `rollup-plugin-typescript2`, but [the user's `tsconfig` overrides](https://github.com/developit/microbundle/blob/a6eb6e75b0bea09ddcb127689cde77b4e75f86f2/src/index.js#L507) with a deep merge (with the small exception of `module` and `target` -- those Microbundle does enforce). 

rpt2 has not been updated for TS 4.1, so until it does, you can't user newer features.

developit:
I think this may be superceded by #741.

PR #764, opened 2020-12-12
Add support for generating inline sourcemaps

bakerkretzmar:
This PR adds support for generating inline sourcemaps with `microbundle --sourcemap inline`.

Although usage of inline sourcemaps isn't common, they're handy for debugging purposes in development when including sourcemaps in a separate file isn't an option.

We use microbundle to build [Ziggy](https://github.com/tighten/ziggy), and the way most people use the package is to inject our minified script right into their templates. This doesn't require them to build or publish any separate files, but it also means we _can't_ add or publish any files, like sourcemaps. Inlining them would allow us to provide that debugging benefit in development without adding any additional setup from users.

Inlined sourcemaps are base-64 encoded and then included directly in the build output where the source map URL would usually go:

```js
var r=function(){try{var r=arguments;return Promise.resolve([].slice.call(r).reduce(function(r,e){return r+e},0))}catch(r){return Promise.reject(r)}};export default function(){try{var e=arguments,t=[].slice.call(e);return Promise.resolve(r.apply(void 0,t)).then(function(e){return Promise.resolve(r.apply(void 0,t)).then(function(r){return[e,r]})})}catch(r){return Promise.reject(r)}}
//# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIjozLCJmaWxlIjoiaW5saW5lLXNvdXJjZS1tYXAuZXNtLmpzIiwic291cmNlcyI6WyIuLi9zcmMvdHdvLmpzIiwiLi4vc3JjL2luZGV4LmpzIl0sInNvdXJjZXNDb250ZW50IjpbImV4cG9ydCBhc3luYyBmdW5jdGlvbiB0d28oLi4uYXJncykge1xuXHRyZXR1cm4gYXJncy5yZWR1Y2UoKHRvdGFsLCB2YWx1ZSkgPT4gdG90YWwgKyB2YWx1ZSwgMCk7XG59XG4iLCJpbXBvcnQgeyB0d28gfSBmcm9tICcuL3R3byc7XG5cbmV4cG9ydCBkZWZhdWx0IGFzeW5jIGZ1bmN0aW9uKC4uLmFyZ3MpIHtcblx0cmV0dXJuIFthd2FpdCB0d28oLi4uYXJncyksIGF3YWl0IHR3byguLi5hcmdzKV07XG59XG4iXSwibmFtZXMiOlsidHdvIiwicmVkdWNlIiwidG90YWwiLCJ2YWx1ZSIsImFyZ3MiXSwibWFwcGluZ3MiOiJJQUFzQkEsaUNBQ3JCLHVCQUFPLGlCQUFLQyxPQUFPLFNBQUNDLEVBQU9DLFVBQVVELEVBQVFDLEdBQU8sSUFEckQsa0ZDRWlDQywwQ0FDbEJKLGVBQU9JLDRDQUFhSixlQUFPSSxxQkFBekMsTUFBTyxVQURSIn0=
```

I've also added snapshot tests that confirm everything works correctly. The fixture test package is a copy of the existing `basic` one with a slightly different `package.json`.

I removed the default `--sourcemap` option of `true` in `prog.js` because it causes the value of this option to always be cast to a boolean, which makes passing `'inline'` impossible. If the `--sourcemap` option isn't passed it still defaults to `true`, so removing the default value doesn't affect microbundle's behaviour at all.


src/index.js
```diff
@@ -57,7 +57,7 @@ export default async function microbundle(inputOptions) {
 	options.name = finalName;
 	options.pkg.name = pkgName;
 
-	if (options.sourcemap !== false) {
+	if (options.sourcemap !== false && options.sourcemap !== 'inline') {
 		options.sourcemap = true;
 	}
```
developit, line 62: Might be good to have a warning for this, so folks don't use the option thinking it's a way to make output cleaner (fewer files).

```suggestion
	if (options.sourcemap === 'inline') {
		console.log(`Warning: inline sourcemaps should only be used for debugging purposes.`);
		options.sourcemap = true;
	} else if (options.sourcemap !== false) {
		options.sourcemap = true;
	}
```

bakerkretzmar: Good idea 👍🏻

developit (approved):

PR #769, opened 2020-12-18
Add --css inline option

### LINKED ISSUE #744: Inline CSS file as string - CJS bundle simply converts this to `void 0`
Bit of a weird use case, but I need to inline a CSS file as a string within a bundle. 

I can do this just fine with UMD, ES Module, and Modern outputs, but where they contain a string representation of the CSS file, the CJS output just has `void 0` in its place. Bug, or is this a limitation of the CJS output in some way?

```
import stylesheet from "./css/styles.css";
console.log(stylesheet);
```

While `console.log()` is certainly not the "real" use case, it acts the same as the real use case and is simpler to display with. 

This is also with the `--css-modules` flag having been set to `false`.

developit:
Fixes #744

### DIFF
--- a/.changeset/rude-keys-eat.md
+++ b/.changeset/rude-keys-eat.md
@@ -1,0 +1,5 @@
+---
+"microbundle": minor
+---
+
+Add `--css inline` option. The default CSS output for all formats is now external files (as it was supposed to be).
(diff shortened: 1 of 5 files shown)


stipsan (approved):
LGTM 👍

rschristian:
This is amazing!! It lets us easily support CSS-in-JS users who might not have a css-loader in their projects but still want to use our styles.

PR #738, opened 2020-10-13
chore: upgrade all dependencies

wardpeet:
I've revisited the upgrade dependencies PR. This one has less changes and it's easier to follow.

Updated majors:
| package                     | old    | new    | changelog                                           |
|-----------------------------|--------|--------|-----------------------------------------------------|
| rollup                      | 1.32.1 | 2.29.0 | <ul><li>[http://github.com/rollup/rollup/releases/tag/v2.0.0](http://github.com/rollup/rollup/releases/tag/v2.0.0)</li></ul> |
| @rollup/plugin-commonjs     | 13.0.0 | 15.1.0 | <ul><li>[https://github.com/rollup/plugins/blob/master/packages/commonjs/CHANGELOG.md#v1400](https://github.com/rollup/plugins/blob/master/packages/commonjs/CHANGELOG.md#v1400)</li><li>[https://github.com/rollup/plugins/blob/master/packages/commonjs/CHANGELOG.md#v1500](https://github.com/rollup/plugins/blob/master/packages/commonjs/CHANGELOG.md#v1500)</li></ul> |
| @rollup/plugin-node-resolve | 6.1.0  | 9.0.0  | <ul><li>[https://github.com/rollup/plugins/blob/master/packages/node-resolve/CHANGELOG.md#v700](https://github.com/rollup/plugins/blob/master/packages/node-resolve/CHANGELOG.md#v700)</li><li>[https://github.com/rollup/plugins/blob/master/packages/node-resolve/CHANGELOG.md#v800](https://github.com/rollup/plugins/blob/master/packages/node-resolve/CHANGELOG.md#v800)</li><li>[https://github.com/rollup/plugins/blob/master/packages/node-resolve/CHANGELOG.md#v900](https://github.com/rollup/plugins/blob/master/packages/node-resolve/CHANGELOG.md#v900)</li></ul> |
| camelcase                   | 5.3.1  | 6.1.0  | <ul><li>[https://github.com/sindresorhus/camelcase/releases/tag/v6.0.0](https://github.com/sindresorhus/camelcase/releases/tag/v6.0.0)</li></ul> |
| kleur                       | 3.0.3  | 4.1.3  | <ul><li>[https://github.com/lukeed/kleur/tree/v4.0.0](https://github.com/lukeed/kleur/tree/v4.0.0)</li></ul> |
[...]

PR #770, opened 2020-12-19
restore CSS sourcemap support

stipsan:
Seems like a regression happened when we merged in #738, it looks like it may have [slipped through](https://github.com/developit/microbundle/commit/bd5d15e17c882f2090f519d342dd89e694456ab8#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556L460) while dealing with merge conflicts.

Since the CSS sourcemap support isn't in any release yet we should be able to reuse the pending change set instead of making a new one for this regression fix: https://github.com/developit/microbundle/blob/master/.changeset/hip-spies-happen.md


src/index.js
```diff
@@ -447,6 +447,7 @@ function createConfig(options, entry, format, writeMeta) {
 						inject: false,
 						extract: options.css !== 'inline',
 						minimize: options.compress,
+						sourceMap: options.sourcemap,
```
developit, line 450: Think it'd be okay if we disabled CSS sourcemaps for inlined CSS? They increase bundle size a ton because of the Base64 encoding.

```suggestion
						sourceMap: options.sourcemap && options.css !== 'inline',
```

stipsan: good point, deffo want to avoid that

developit (approved):

developit:
aces.

PR #777, opened 2021-01-06
fix: Extra CSS files generated

### LINKED ISSUE #776: Bug: Microbundle generates 'x.modern.css' and 'x.module.css' files
As the title indicates, as of v13, Microbundle generates `x.modern.css` and `x.module.css` (and their source maps). The UMD build, however, seems to be safe from this as there is not a `x.umd.css`.

Here's a repo that one can use to reproduce: https://github.com/rschristian/microbundle-css-output

I've pushed a build to the repo for easier viewing.

rschristian:
## Summary

Fixes #776 

## Description

A bug with #769 meant that for each JS file output format (excluding UMD) a matching CSS file would be generated, i.e., `index.modern.css` and `index.module.css` (not a CSS module).

I corrected this by [restoring the old logic used](https://github.com/developit/microbundle/commit/967f8d532785aa7bf8636c5a759759a3e72dcf56#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556L448), which was relying on `writeMeta`.

### DIFF
--- a/.changeset/modern-students-give.md
+++ b/.changeset/modern-students-give.md
@@ -1,0 +1,5 @@
+---
+'microbundle': patch
+---
+
+Fixing a bug that would cause a CSS file to be generated to match each JS build output
--- a/src/index.js
+++ b/src/index.js
@@ -458,7 +458,7 @@ function createConfig(options, entry, format, writeMeta) {
 						// only write out CSS for the first bundle (avoids pointless extra files):
 						inject: false,
 						extract:
-              !!writeMeta &&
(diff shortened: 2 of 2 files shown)


developit:
Seems reasonable to me - this doesn't create any incorrect references to the CSS files right? I don't think we provide a way to get the filename so my guess is that it's fine.

developit (approved):

rschristian:
I don't believe there are any references, no.

developit (approved):

Issue #775, opened 2021-01-01
--format=modern still converts modern JS to old-style JS?


Pomax:
I was trying to figure out why my code was doing the wrong thing, and ran into the fact that using `--format=modern` still seems to yield legacy JS:

```
> microbundle watch -i ./src/react-app/ui.js -o ./public/preact-ui.js --format=modern --no-sourcemap --no-compress
```

turned the following:

```
class MJUI extends Component {
    constructor(props) {
      // ...
    }
  
  renderGames(state) {
    return state.games.map((g) => {
      if (g.finished) return;

      let joinOrStart = async () => {
        // Note that these two operations are mutually
        // exclusive, because owner's can't join, and
        // non-owners can't start a game:
        this.server.game.join(g.name);
        this.server.game.start(g.name);
      };

      // ...
    });
  }

  // ...
}
```

into:

```
  // ...

  renderGames(state) {
    var _this = this;

    return state.games.map(g => {
      if (g.finished) return;

      let joinOrStart = async function joinOrStart() {
        // Note that these two operations are mutually
        // exclusive, because owner's can't join, and
        // non-owners can't start a game:
        _this.server.game.join(g.name);

        _this.server.game.start(g.name);
      };

  // ...
```

Note that the filename has the `.modern.js` suffix, so something feels rather off?

developit:
This is actually correct - the original code causes an unrecoverable error when renderGames() is called in Safari 10 and 11 (there is a bug relating to async arrow functions in classes). Microbundle uses preset-env's `bugfixes` option, which converts the async arrow function to an async function, because that's far faster and cheaper than transpiling the async/await to Regenerator.

Pomax:
Hm, how do I tell it that I couldn't care less about browsers that don't support the modern web?

Pomax:
(and happy new year, of course =)

developit:
For that you'd want to ignore the "modern" output format and use the "esm" format, then specify your support target using [Browserslist](https://github.com/browserslist/browserslist#packagejson):

```
// package.json
{
  "browserslist": [
    "last 2 years",
    "node >= 12"
  ]
}
```

The modern output format is pretty specifically designed to generate ES2017 code because that offers 95-96% browser support, whereas newer syntax levels are sub-90%.

Pomax:
Hmm, I made those changes but I can't seem to get it to leave the code alone. I figured I'd set it purely chrome and firefox, to see if that'd yield the least amount of rewriting:

```json
[...]

PR #778, opened 2021-01-06
docs: Swapping out references of ES format to ESM

rschristian:
## Summary 

Addresses the footer of https://github.com/developit/microbundle/issues/775#issuecomment-753371904

Replaces "`es`" in the output formats section of the ReadMe + CLI's help with "`esm`". The format is otherwise always referred to as "ESM" or "ES Modules" so I agree that's a fair change to make. 

Note: There's a few unrelated changes caused by the prettier pre-commit hook. I'd normally revert the unrelated changes but I figure the hook is there to be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,11 +173,11 @@ Importing CSS files is supported via `import "./foo.css"`. By default, generated
 
 ```js
 // with the default external CSS:
-import './foo.css';  // generates a minified .css file in the output directory
+import './foo.css'; // generates a minified .css file in the output directory
 
 // with `microbundle --css inline`:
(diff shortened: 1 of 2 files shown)


developit (approved):
LGTM! We have these aliased anyway so its nice to standardize. I believe Rollup also allows both, though [they have `es` as the default](https://rollupjs.org/guide/en/#outputformat).

rschristian:
Ah, that makes sense! I noticed that `esm` gets converted to `es` after checking that both were supported but wasn't sure of the reason for it.

PR #781, opened 2021-01-07
docs: Updating ReadMe to use correct path for "exports" field

rschristian:
## Summary

Corrects the `"exports"` field in one of the ReadMe sections to follow [the spec about exports' paths](https://nodejs.org/api/packages.html#packages_exports).

I did only update the `"exports"` field and left the others alone. Stylistically I'm not sure whether consistency is better or not here. The [block in the "Modern Mode" section](https://github.com/developit/microbundle#-modern-mode-) adds the `./` prefix to all fields but in other sections, where `"exports"` is omitted, `./` is not present.

Not sure what others' opinions are on the stylistic portion, but at least the field itself is correct to copy/paste with this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,15 +38,15 @@
 
 ```js
 {
-  "name": "foo",                   // your package name
-  "source": "src/foo.js",          // your source code
-  "main": "dist/foo.js",           // where to generate the CommonJS/Node bundle
-  "exports": "dist/foo.modern.js", // path to the modern output (see below)
-  "module": "dist/foo.module.js",  // where to generate the ESM bundle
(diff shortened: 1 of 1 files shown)


ludofischer:
I _think_ `./` should be fine everywhere, since other examples in the README use `./` everywhere (after 'This is enabled by default. All you have to do...')

rschristian:
@ludofischer Well, it's a mix. As I said, when `exports` is present in the example, then all fields use the `./` prefix. When it's not, they don't. There's a couple instances of each in the ReadMe. 

It makes no difference functionally, yes. Just a bit more busy looking then it needs to be, though that gripe is largely with the `exports` field spec...

rschristian:
Closing, already fixed

PR #784, opened 2021-01-07
feat: Support customizing the modern output via "exports"

### LINKED ISSUE #783: Feature: Customize modern output file name using the "exports" field
Currently, users are not able to configure the modern output file name by using the `"exports"` field. For example, say the following `package.json` was being used:

```
{
	"main": "./dist/index.js",
	"module": "./dist/index.example.js",
	"exports": "./dist/index.example-two.js",
}
```

Microbundle would return `index.js`, `index.example.js`, and `index.modern.js`. The fact that the user has it as `index.example-two.js` would be ignored.

This, at the moment, is only [customizable using `"esmodules"`](https://github.com/developit/microbundle/blob/master/src/index.js#L282-L285), though I don't see why `exports` shouldn't work too. 

I'm not sure how often such a thing is used but I ran across it whilst running some tests. I'll write up a quick PR here. So long as `"exports"` is a string I think we should be safe to just use that value?

rschristian:
## Summary

Closes #783 

Allows users to customize modern output file name with `exports` like they can with `esmodules`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -285,7 +285,7 @@ function getMain({ options, entry, format }) {
 		mainNoExtension,
 	);
 	mainsByFormat.modern = replaceName(
-		(pkg.exports && walk(exports.exports)) ||
+		(pkg.exports && walk(pkg.exports)) ||
 			(pkg.syntax && pkg.syntax.esmodules) ||
 			pkg.esmodule ||
 			'x.modern.js',
--- a/src/index.js
+++ b/src/index.js
@@ -251,6 +251,11 @@ function replaceName(filename, name) {
 	);
 }
 
+function walk(exports) {
+	if (typeof exports === 'string') return exports;
+	return walk(exports['.'] || exports.import || exports.module);
+}
+
 function getMain({ options, entry, format }) {
 	const { pkg } = options;
 	const pkgMain = options['pkg-main'];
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -280,7 +280,10 @@ function getMain({ options, entry, format }) {
 		mainNoExtension,
 	);
 	mainsByFormat.modern = replaceName(
-		(pkg.syntax && pkg.syntax.esmodules) || pkg.esmodule || 'x.modern.js',
+		(typeof pkg.exports === 'string' && pkg.exports) ||
+			(pkg.syntax && pkg.syntax.esmodules) ||
```
developit, line 284: I think this proposal was superceded by pkg.exports and can be removed

src/index.js
```diff
@@ -280,7 +280,10 @@ function getMain({ options, entry, format }) {
 		mainNoExtension,
 	);
 	mainsByFormat.modern = replaceName(
-		(pkg.syntax && pkg.syntax.esmodules) || pkg.esmodule || 'x.modern.js',
+		(typeof pkg.exports === 'string' && pkg.exports) ||
```
developit, line 283: This works for the main case. I'm wondering - maybe we could walk the exports object to find matching `import` entries?

```
function walk(exports) {
  if (typeof exports === 'string') return exports;
  return walk(exports['.'] || exports.import || exports.module);
}

walk(pkg.exports)
```

src/index.js
rschristian: Is anyone using that? I've never seen it in the wild, but it potentially would be a breaking change, no?

src/index.js
rschristian: Certainly could. I started to think about this but had reservations on whether or not users would expect all keys (or at least the ones in the spec?) to be supported. If they have a `"require"`, would they then expect a matching output file? Not sure.

src/index.js
developit: Ah, I didn't notice it was previously supported here.

developit (approved):

developit (approved):

Issue #531, opened 2019-12-02
why esmodule set to false?


hardfist:
```
export const a = 10;
export const b = 10;
```
will generates following in microbundle, which missing `__esModule` flag
```
exports.a=10,exports.b=10;
//# sourceMappingURL=index.js.map
```
while using rollup generates the folloing code,which seems more accurate
```
'use strict';

Object.defineProperty(exports, '__esModule', { value: true });

const a = 10;
const b = 10;

exports.a = a;
exports.b = b;
```

developit:
The CommonJS output is CommonJS, not ES Modules. ES Modules consumers should use the ES Modules output.

hardfist:
It works for project using webpack or rollup, but seems doesn't work for node typescript project which only use tsc and doesn't consume module field

DaniAkash:
@hardfist Indeed!
I followed this [workaround](https://github.com/developit/microbundle/issues/306#issuecomment-459103476) but this causes my library to not work in TypeScript environments...

developit:
FWIW Microbundle used to detect name-only and default-only exports and remove `__esModule` only in those cases. It's possible we need to provide an option to bring that behavior back.

hardfist:
> FWIW Microbundle used to detect name-only and default-only exports and remove `__esModule` only in those cases. It's possible we need to provide an option to bring that behavior back.

why remove `__emModule`

huozhi:
consider make `esModule` as an option or param in CLI?

wardpeet:
@huozhi Could you explain a little bit more about the use case? Are you targeting node 14?

huozhi:
@wardpeet the purpose is to use esmodule syntax `import` with cjs bundles. usually cjs assets define a property `__esModule` to work compatiable with interpolate import. but pure cjs (disable esModule config) will erase property `__esModule` that breaks import on cjs files.

```js
import defaultExportLib from 'lib' // this won't import the default export variable from 'lib' when use pure commonjs format
```

typescript has an option `esModuleInterop` to enable this module transform behavior, which define __esModule property on the target modules. the generated assets will look like following:

```js
function __export(m) {
    for (var p in m) if (!exports.hasOwnProperty(p)) exports[p] = m[p];
}
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
Object.defineProperty(exports, "__esModule", { value: true });
__export(require("./use-swr"));
[...]

PR #785, opened 2021-01-13
enhance: support tsconfig esmodule interop

huozhi:
### description

related to #531
when there's multiple exports in the source file while `esModuleInterop` specified as `true` in tsconfig compilerOptions, let rollup add `__esModule: {value: true}` mark to the output.

it will benefits the esmodule usage with the dist file when the dist is resolved to commonjs assets. the reason could be:
* only main field is existed in package.json
* due to some reason cjs assets is resolved instead of esmodule assets.

### changes
* respect `esModuleInterop` in tsconfig (for typescript source only)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -400,6 +400,19 @@ function createConfig(options, entry, format, writeMeta) {
 	const absMain = resolve(options.cwd, getMain({ options, entry, format }));
 	const outputDir = dirname(absMain);
 	const outputEntryFileName = basename(absMain);
+	let tsconfigPath;
+	let tsconfigOptions = {};
+	let ts;
+	if (useTypescript) {
+		tsconfigPath = resolve(
+			options.tsconfig || resolve(options.cwd, 'tsconfig.json'),
+		);
+		ts = require(resolveFrom.silent(options.cwd, 'typescript') || 'typescript');
+		const tsconfigJSON = ts.readConfigFile(tsconfigPath, ts.sys.readFile)
+			.config;
+		tsconfigOptions = ts.parseJsonConfigFileContent(tsconfigJSON, ts.sys, './')
+			.options;
+	}
 
 	let config = {
 		/** @type {import('rollup').InputOptions} */
@@ -480,10 +493,7 @@ function createConfig(options, entry, format, writeMeta) {
 					},
 					(useTypescript || emitDeclaration) &&
 						typescript({
-							typescript: require(resolveFrom.silent(
-								options.cwd,
-								'typescript',
-							) || 'typescript'),
+							typescript: ts,
 							cacheRoot: `./node_modules/.cache/.rts2_cache_${format}`,
 							useTsconfigDeclarationDir: true,
 							tsconfigDefaults: {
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -400,6 +400,19 @@ function createConfig(options, entry, format, writeMeta) {
 	const absMain = resolve(options.cwd, getMain({ options, entry, format }));
 	const outputDir = dirname(absMain);
 	const outputEntryFileName = basename(absMain);
+	let tsconfigPath;
+	let tsconfigOptions = {};
+	let ts;
+	if (useTypescript) {
+		tsconfigPath = resolve(
+			options.tsconfig || resolve(options.cwd, 'tsconfig.json'),
+		);
```
rschristian, line 409: You don't need this. `rollup-plugin-typescript2` will already try to load `./tsconfig.json` https://www.npmjs.com/package/rollup-plugin-typescript2#plugin-options

src/index.js
```diff
@@ -494,6 +504,7 @@ function createConfig(options, entry, format, writeMeta) {
 									emitDeclarationOnly: options.generateTypes && !useTypescript,
 									declarationDir: getDeclarationDir({ options, pkg }),
 									jsx: 'preserve',
+									esModuleInterop: !!tsconfigOptions.esModuleInterop,
```
rschristian, line 507: I don't really understand the reason for this? You're setting a default to come from the user's TSConfig, but the user's TSConfig overrides these defaults.

src/index.js
huozhi: the approach here is aimed to get the tsconfig json object, then I could use it for rollup output options.
if the interop is enabled, we want to add the __esModule mark for cjs output.

src/index.js
huozhi: good catch, this is kinda duplicated, I've removed it

src/index.js
rschristian: Gotcha, that's fair. Sorry, misunderstood.

src/index.js
```diff
@@ -503,7 +513,7 @@ function createConfig(options, entry, format, writeMeta) {
 								},
 								files: options.entries,
 							},
-							tsconfig: options.tsconfig,
+							tsconfig: tsconfigPath,
```
rschristian, line 516:
```suggestion
							tsconfig: options.tsconfig,
```

This is the default behaviour with rpt2.

src/index.js
```diff
@@ -400,6 +400,19 @@ function createConfig(options, entry, format, writeMeta) {
...
 	const outputEntryFileName = basename(absMain);
+	let tsconfigPath;
+	let tsconfigOptions = {};
+	let ts;
+	if (useTypescript) {
+		tsconfigPath = resolve(
+			options.tsconfig || resolve(options.cwd, 'tsconfig.json'),
+		);
+		ts = require(resolveFrom.silent(options.cwd, 'typescript') || 'typescript');
+		const tsconfigJSON = ts.readConfigFile(tsconfigPath, ts.sys.readFile)
```
rschristian, line 411: You'll need to check to ensure the tsconfig actually exists. I think a user can get by using Microbundle without one, thanks to the tsconfig defaults

rschristian: This will also be relevant down in the `esModule` change too

wardpeet:
I wonder if we should add interop for non esm modules and look at https://github.com/rollup/rollup/pull/3710 perhaps instead?

huozhi:
@wardpeet ty, I'll look for other ways to workaround

PR #794, opened 2021-01-31
docs: use "exports" key in all examples

ludofischer:
I think having two fields in the examples is confusing, especially since I could not find recent sources recommending `esmodule` any more.
Related to https://github.com/developit/microbundle/pull/781

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,10 @@ Microbundle uses the fields from your `package.json` to figure out where it shou
   "main": "dist/foo.js",            // CommonJS bundle
   "umd:main": "dist/foo.umd.js",    // UMD bundle
   "module": "dist/foo.m.js",        // ES Modules bundle
-  "exports": "dist/foo.modern.js", // Modern ES Modules bundle
+  "exports": {
+    "require": "./dist/foo.js",  // CommonJS bundle
+    "default": "./dist/foo.modern.js", // Modern ES Modules bundle
+  }
   "types": "dist/foo.d.ts"          // TypeScript typings directory
 }
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -204,7 +204,7 @@ Microbundle uses the fields from your `package.json` to figure out where it shou
   "main": "dist/foo.js",            // CommonJS bundle
   "umd:main": "dist/foo.umd.js",    // UMD bundle
   "module": "dist/foo.m.js",        // ES Modules bundle
-  "esmodule": "dist/foo.modern.js", // Modern bundle
+  "exports": "dist/foo.modern.js", // Modern ES Modules bundle
```
developit, line 207:
```suggestion
  "exports": {
    "require": "./dist/foo.js",  // CommonJS bundle
    "default": "./dist/foo.modern.js", // Modern ES Modules bundle
  }
```

developit (commented):
This looks good, but I need to check if it actually works without using a `.mjs` file extension for the `"import"` field value. If not, it might be better to recommend using `.cjs` for the CommonJS and UMD variants, with `"type":"module"` in the package.json.

README.md
ludofischer: So main is still required for Node.js < 12.0, and the `exports.require` is there to cover CommonJS support for Node.js 12.0+, did I get this right?

ludofischer:
I've had another go at this to at least make the docs internally consistent. But looking at issues like https://github.com/rollup/plugins/issues/695, I get the feeling that there is not 'one' way to organize the `package.json` field that will guarantee every tool will pick up the correct exports.

developit (approved):

PR #798, opened 2021-02-02
fix(transform-fast-rest): rest parameters references with closures

### LINKED ISSUE #797: [Transpiling] Rest parameters doesn't work with non-arrow function closures
Hello there!

I've encountered a bug with the transpiling of the rest parameters.

Pratically the transpiler with something like this:
```js
export function brokenArgsRest(fn, ...args) {
  return function () {
    fn(...args);
  };
}
```

Will treat the arguments object in the wrong way:
```js
function brokenArgsRest(fn) {
  return function () {
    fn.apply(void 0, [].slice.call(arguments, 1));
  };
}
```

I've created a minimal repro here --> https://github.com/gdorsi/microbundle-args-rest-issue

gdorsi:
closes #797

### DIFF
--- a/src/lib/transform-fast-rest.js
+++ b/src/lib/transform-fast-rest.js
@@ -46,8 +46,7 @@ export default function fastRestTransform({ template, types: t }) {
 				if (
 					binding.constant &&
 					binding.referencePaths.length === 1 &&
-					// arguments access requires the same function scope
-					getParentFunction(binding.referencePaths[0], t) === func
+					sameArgumentsObject(binding.referencePaths[0], func, t)
 				) {
 					// one usage, never assigned - replace usage inline
 					binding.referencePaths[0].replaceWith(sliced);
@@ -94,10 +93,16 @@ export default function fastRestTransform({ template, types: t }) {
 	};
 }
 
-function getParentFunction(node, t) {
+function sameArgumentsObject(node, func, t) {
 	while ((node = node.parentPath)) {
-		if (t.isFunction(node)) {
-			return node;
+		if (node === func) {
+			return true;
+		}
+
+		if (t.isFunction(node) && !t.isArrowFunctionExpression(node)) {
+			return false;
 		}
 	}
+
+	return false;
 }
--- a/src/lib/transform-fast-rest.js
+++ b/src/lib/transform-fast-rest.js
@@ -43,7 +43,12 @@ export default function fastRestTransform({ template, types: t }) {
 					func.arrowFunctionToExpression();
 				}
 
-				if (binding.constant && binding.referencePaths.length === 1) {
+				if (
+					binding.constant &&
+					binding.referencePaths.length === 1 &&
+					// arguments access requires the same function scope
+					getParentFunction(binding.referencePaths[0], t) === func
+				) {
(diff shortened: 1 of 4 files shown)


src/lib/transform-fast-rest.js
```diff
@@ -43,7 +43,12 @@ export default function fastRestTransform({ template, types: t }) {
 					func.arrowFunctionToExpression();
 				}
 
-				if (binding.constant && binding.referencePaths.length === 1) {
+				if (
+					binding.constant &&
+					binding.referencePaths.length === 1 &&
+					// arguments access requires the same function scope
+					getParentFunction(binding.referencePaths[0], t) === func
```
developit, line 50: nit: there's a builtin for this (though see my other comment for why you might actually have the better approach here!)

```suggestion
					binding.referencePaths[0].getFunctionParent() === func
```

src/lib/transform-fast-rest.js
```diff
@@ -88,3 +93,11 @@ export default function fastRestTransform({ template, types: t }) {
 		},
 	};
 }
+
+function getParentFunction(node, t) {
+	while ((node = node.parentPath)) {
+		if (t.isFunction(node)) {
```
developit, line 99: Arrow functions can access `arguments` from the enclosing scope, so we could allow them in the tree here:

```suggestion
		if (t.isFunction(node) && !t.isArrowFunctionExpression(node)) {
```

developit (commented):
This is great! Curious to get your thoughts on the arrow function bit, otherwise this looks ready to go.

src/lib/transform-fast-rest.js
gdorsi: aww, that's nice!

src/lib/transform-fast-rest.js
gdorsi: I thought about that, but if the rest parameters transformation is applied with the same criteria of the arrow function transformation there would be no gain in terms of bundlesize.

Instead if the `[].slice.call` is moved in the root function scope it would be cached in some cases.

gdorsi: I've made the change and without the arrow function check the bundle size of the test case seems smaller:

With the arrow function check:
```
157 B: parameters-rest-closure.js.gz
115 B: parameters-rest-closure.js.br
165 B: parameters-rest-closure.esm.js.gz
124 B: parameters-rest-closure.esm.js.br
247 B: parameters-rest-closure.umd.js.gz
189 B: parameters-rest-closure.umd.js.br
```

With getFunctionParent
```
146 B: parameters-rest-closure.js.gz
106 B: parameters-rest-closure.js.br
153 B: parameters-rest-closure.esm.js.gz
113 B: parameters-rest-closure.esm.js.br
238 B: parameters-rest-closure.umd.js.gz
177 B: parameters-rest-closure.umd.js.br
```

developit (approved):
LGTM!

src/lib/transform-fast-rest.js
developit: makes sense from a Gzip perspective (repetition is essentially free), just there's a runtime cost associated with materializing `arguments` via slice() where it wouldn't otherwise be necessary.

PR #799, opened 2021-02-02
fix(cli): corrects the --sourcemap=false option parsing

### LINKED ISSUE #773: --sourcemap=false does nothing?
I'm running the following command:
```
> npx microbundle -i ./src/react-app/ui.js -o ./public/preact-ui.js --format=esm --sourcemap=false 
```
and while the `--format` flag works (only the ESM version gets generated) the `--sourcemap` flag seems to do nothing: there's still a `preact-ui.esm.js.map` file getting generated, with a comment that points to that file in the `preact-ui.esm.js` file.

gdorsi:
closes #773

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -61,6 +61,8 @@ export default async function microbundle(inputOptions) {
 		console.log(
 			'Warning: inline sourcemaps should only be used for debugging purposes.',
 		);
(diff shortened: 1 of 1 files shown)


developit (approved):

developit:
heh - happy to have this one fixed.

Issue #801, opened 2021-02-04
package.json "type": "module" / [ERR_REQUIRE_ESM]: Must use import to load ES Module


aheissenberger:
**Problem:**
`microbundle` does not correctly bundling packages where the package type is `"type": "module"` in `package.json`

when importing `const appfaker = require('@ekzemplo/common')` the bundled package into a project which is using `"type": "commonjs"` this error id thrown:
```
node:internal/modules/cjs/loader:1125
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /TST/ekzemplo/packages/ekzemplo-common/dist/index.js
require() of ES modules is not supported.
require() of /TST/ekzemplo/packages/ekzemplo-common/dist/index.js from /TST/ekzemplo/packages/ekzemplo-apievents/src/server.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /TST/ekzemplo/packages/ekzemplo-common/package.json.
...
```

this is the `package.json` of the imported package created with `microbundle`

```json
{
  "name": "@ekzemplo/common",
  "version": "0.0.0",
  "source": "./src/index.js",
  "main": "./dist/index.js",
  "module": "./dist/index.module.mjs",
  "unpkg": "./dist/index.umd.js",
  "exports": {
    "import": "./dist/index.module.mjs",
    "require": "./dist/index.js"
  },
  "type": "module",
  "type":"commonjs",
  "license": "MIT",
  "scripts": {
    "start": "microbundle watch",
    "build": "microbundle build"
  },
  "dependencies": {
    "faker": "^5.3.1"
  }
}
```

**manual fix**
I was able to fix the problem by renaming the file `./dist/index.js` to `./dist/index.cjs` and adapting `main` and `require` in the `package.json` after building:
```json
{
  "name": "@ekzemplo/common",
  "version": "0.0.0",
  "source": "./src/index.js",
  "main": "./dist/index.cjs",
  "module": "./dist/index.module.mjs",
  "unpkg": "./dist/index.umd.js",
  "exports": {
    "import": "./dist/index.module.mjs",
    "require": "./dist/index.cjs"
  },
  "type": "module",
  "type":"commonjs",
  "license": "MIT",
  "scripts": {
[...]

PR #802, opened 2021-02-04
fix default extension to cjs for package.json "type":"module" 

aheissenberger:
The old implementation will not allow to use a filename with the extension ".cjs" in the "main" field and does not create a Common JS file with a ".cjs" ending.

If a package.json contains "type":"module"  all files with ".js" will throw an error if they are imported with an `require` function. 

This fix:
* will strip `.cjs` when creating the base name
* switch from default extension `.js` to `.cjs` when `"type":"module"` exists in `package.json`
* update documentation to highlight the problem

This solves #801

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,11 +173,11 @@ Importing CSS files is supported via `import "./foo.css"`. By default, generated
 
 ```js
 // with the default external CSS:
-import './foo.css';  // generates a minified .css file in the output directory
+import './foo.css'; // generates a minified .css file in the output directory
 
 // with `microbundle --css inline`:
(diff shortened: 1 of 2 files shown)


developit:
Hiya - sorry for the delay on this PR, I'm actually thinking we might be better off detecting this and warning, rather than relying on docs.

developit (approved):

JounQin:
@developit There is no changeset generated for this PR.

jeffryang24:
Hmm.. Can't wait this change to be published... :bow:

developit:
This is published.

PR #823, opened 2021-03-28
Fix/css module types

rschristian:
#768 added an ambient type declaration for CSS Modules, however, this was not included in the list of of files to be published, so it never got to users

I added quick note in the ReadMe for using this, will probably help a few people.


developit (approved):
LGTM!

PR #830, opened 2021-04-06
fix: add generateTypes cli option, check false value correctly

### LINKED ISSUE #829: `generateTypes` is not available in cli, and `generateTypes: false` is not woking
https://github.com/developit/microbundle/blob/master/src/index.js#L369

I'll raise a PR to fix it.

JounQin:
close #829

### DIFF
--- a/.changeset/cold-moose-cough.md
+++ b/.changeset/cold-moose-cough.md
@@ -1,0 +1,5 @@
+---
+"microbundle": patch
+---
+
+fix: add generateTypes cli option, check false value correctly
--- a/README.md
+++ b/README.md
@@ -175,11 +175,11 @@ Importing CSS files is supported via `import "./foo.css"`. By default, generated
 
 ```js
 // with the default external CSS:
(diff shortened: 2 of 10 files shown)


JounQin:
@developit Any time to review?

rschristian:
What's the use case for this? 

As `types` and `typings` are the only standardized keys for TS types, I can't see why you'd want to generate types without having those keys present? Right now I'd say it would make more sense to go the other way, and remove `generateTypes`, though perhaps I'm not seeing a certain use case?

JounQin:
@rschristian We want to write the `types`file manually.

rschristian:
Ah, yeah, that makes sense.

JounQin:
friendly ping @developit

JounQin:
friendly ping @developit

developit (approved):

developit:
Ran into this exact issue yesterday, and your PR seems to be the right fix! Thanks.

PR #840, opened 2021-05-22
Fix/sourcemap cli

### LINKED ISSUE #837: --sourcemap false is not working
The Readme mentions `--sourcemap` as a parameter, but setting it to `false` does nothing. `--no-sourcemap` does what I expect here.

I'll have a look at the source & submit a PR to change this when I have some time.

rschristian:
Closes #837


rschristian:
Oops, I was using an out-dated fork. This was already fixed in #773. My bad!

PR #843, opened 2021-05-26
Fix: throw  a warning when no entrypoint is found

### LINKED ISSUE #404: Microbundle should throw an error when no entrypoint is found
When an entrypoint for microbundle isn't found, it happily states that it's completed its job successfully. This is pretty confusing -- it would be great if it printed out some kind of error!

Simple repro:

Put this `package.json` in an empty folder:
```json
{
  "name": "microbundle-repro",
  "version": "1.0.0",
  "source": "foo.ts",
  "main": "dist/foo.js",
  "module": "dist/foo.mjs",
  "unpkg": "dist/foo.umd.js",
  "types": "dist/foo.d.ts",
  "dependencies": {
    "microbundle": "^0.11.0"
  }
}
```

`yarn install && yarn microbundle`:
```
Build "(foldername)" to dist:
Done in 0.98s.
```

G3root:
fixes #404

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,7 +2,7 @@ import fs from 'fs';
 import { resolve, relative, dirname, basename, extname } from 'path';
 import camelCase from 'camelcase';
 import escapeStringRegexp from 'escape-string-regexp';
-import { blue } from 'kleur';
+import { blue, red } from 'kleur';
 import { map, series } from 'asyncro';
 import glob from 'tiny-glob/sync';
 import autoprefixer from 'autoprefixer';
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -130,7 +130,12 @@ export default async function microbundle(inputOptions) {
 	);
 
 	const targetDir = relative(cwd, dirname(options.output)) || '.';
-	const banner = blue(`Build "${options.name}" to ${targetDir}:`);
+	const sourceExist = options.input.length > 0;
+	const banner = sourceExist
+		? blue(`Build "${options.name}" to ${targetDir}:`)
+		: red(
+				`[!] Error: Can't resolve '${options.pkg.source}' in '${options.cwd}'`,
+		  );
```
developit, line 138: This might be confusing for folks who aren't using `"source"`. I am thinking maybe a more generic error message would avoid that:

```suggestion
		: red(`Error: No entry module found for "${options.name}"`);
```

G3root: Can you provide me a example message I'll create a pull request

developit (approved):
Thanks!

PR #853, opened 2021-06-12
Fix exports walk (#852)

### LINKED ISSUE #852: Error while walking package exports
Given the following exports in package.json
```
  "exports": {
    ".": {
      "import": {
        "webpack": "./dist/index.import.js",
        "default": "./dist/index.import.mjs"
      },
      "require": "./dist/index.js",
      "script": "./dist/index.umd.js",
      "default": "./dist/index.import.mjs"
    }
  },
```

fails with: 
```
TypeError: Cannot read property '.' of undefined

    at walk (C:\VSProjects\github\mobx-keystone\node_modules\microbundle\dist\cli.js:810:22)
    at walk (C:\VSProjects\github\mobx-keystone\node_modules\microbundle\dist\cli.js:810:10)
    at walk (C:\VSProjects\github\mobx-keystone\node_modules\microbundle\dist\cli.js:810:10)
    at walk (C:\VSProjects\github\mobx-keystone\node_modules\microbundle\dist\cli.js:810:10)
    at getMain (C:\VSProjects\github\mobx-keystone\node_modules\microbundle\dist\cli.js:837:53)
    at createConfig (C:\VSProjects\github\mobx-keystone\node_modules\microbundle\dist\cli.js:927:45)
    at microbundle (C:\VSProjects\github\mobx-keystone\node_modules\microbundle\dist\cli.js:659:18)
```

Apparently it is first walking inside the ".", then inside "import" and then it fails because "import" is an object without ".", "import" or "module"

It used to work in 0.13.1 but started failing in 0.13.2 and 0.13.3

developit:
Fixes #852

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -260,6 +260,7 @@ function replaceName(filename, name) {
 }
 
 function walk(exports) {
+	if (!exports) return null;
 	if (typeof exports === 'string') return exports;
 	return walk(exports['.'] || exports.import || exports.module);
 }


rschristian (approved):
This does start to make me worry that #784 was a bad idea, due to the amount of possible combinations that we can't reasonably cover, requiring users to fix filenames on their end.

src/index.js
```diff
@@ -260,6 +260,7 @@ function replaceName(filename, name) {
 }
 
 function walk(exports) {
+	if (!exports) return null;
 	if (typeof exports === 'string') return exports;
 	return walk(exports['.'] || exports.import || exports.module);
```
jeffryang24, line 265: Hmm... I found a regression when declaring modern exports using default conditional export, such as:

```json
"exports": {
    ".": {
      "import": "dist/index.js",
      "require": "dist/index.cjs",
      "default": "dist/index.modern.js"
    }
  }
```

The modern module is not being generated in version >= 0.13.2 while it is being generated in v0.13.1. Probably, we can also add `exports.default` inside the `walk` recursion method. wdyt @developit ? :bow: 

```suggestion
	return walk(exports['.'] || exports.import || exports.module || exports.default);
```

jeffryang24: any update regarding this fix? :bow:

developit: Hi @jeffryang24 - no update, just haven't gotten around to it yet.

developit: I think we could add `.default` as you suggest, but really this needs to also be checking the extension based on `pkg.type`, otherwise it will potentially return a CommonJS entry filename for the modern filename.

jeffryang24: I see... got it... Currently, I've reverted back to 0.13.1 to fix this regression issue. Thanks for your explanation... :bow:

ChristophP: @jeffryang24 thanks, I also reverted to 0.13.1 and that makes it work for me for the moment.

PR #858, opened 2021-06-28
Fix: Support booleans in minify options

bouchenoiremarc:
**Context:** Disabling Terser's mangling completely is only possible by setting `mangle` to `false`, but currently Microbundle replaces any non-object `mangle` option by `{}`.

This PR adds a special condition to allow the minify options `compress` and `mangle` to be set as booleans.


developit (approved):
This is awesome, and I have been wanting this feature for a while. Perfect PR!

PR #865, opened 2021-07-15
feat: Expands generateTypes flag to support TS entries

### LINKED ISSUE #863: Clarify the correct way to not generate types
In my project I have
```json
// tsconfig.json
{
	"compilerOptions": {
		"checkJs": true,
		"target": "es5",
		"lib": ["ES2015"],
		"outDir": "dist",
		"skipLibCheck": true,
		"esModuleInterop": true,
		"declaration": true,
    "moduleResolution": "node", // Tell tsc to look in node_modules for modules
    "sourceMap": true,          // Creates *.js.map files
		"strict": true,             // Strict types, eg. prohibits `var x=0; x=null`
    "alwaysStrict": true        // Enable JavaScript's "use strict" mode
	},
	"include": ["src/**/*.ts", "src/**/*.tsx"],
  "exclude": ["../node_modules", "node_modules"]
}
```

For a deploy I do not want to take the time to generate types for the rest of the project in JS (this is a single project in a yarn workspace).

none of the following seem to work:

```
--generateTypes false
--generateTypes=false
--no-generate-types
```

If I use a separate `tsconfig.deploy.json` for this specifig environment and specify "declarations: false"
```json
{
	"compilerOptions": {
		"checkJs": true,
		"target": "es5",
		"lib": ["ES2015"],
		"outDir": "dist",
		"skipLibCheck": true,
		"esModuleInterop": true,
		"declaration": false,
    "moduleResolution": "node", // Tell tsc to look in node_modules for modules
		"strict": true,             // Strict types, eg. prohibits `var x=0; x=null`
    "alwaysStrict": true        // Enable JavaScript's "use strict" mode
	},
	"include": ["src/**/*.ts", "src/**/*.tsx"],
  "exclude": ["../node_modules", "node_modules"]
}
```
I now do not get types, but I also get the following warnings (full script from package.json in first line)
```
$ rm -rf dist && microbundle --tsconfig ./tsconfig.deploy.json --no-sourcemap --generateTypes false --format modern src/index.ts src/data.ts
rpt2: options error TS5069: Option 'declarationDir' cannot be specified without specifying option 'declaration' or option 'composite'.
rpt2: options error TS5069: Option 'declarationDir' cannot be specified without specifying option 'declaration' or option 'composite'.
```

Is there somewhere that microbundle is merging tsconfigs? What is the proper way to prevent type definition generation?

rschristian:
**What kind of change does this PR introduce?**

Feature, expands the `--generateTypes` flag to also support projects with TS entry points.

**Did you add tests for your changes?**

Modified an existing test

**Summary**

Closes #863

The problem with the existing `--generateTypes` behavior is that it is only relevant when used with JS entry points.

`emitDeclaration` is derived from `options.generateTypes` and is only used in the following spot:

https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/src/index.js#L510

Which means that when `useTypescript` is truthy, that option holds no power whatsoever. `useTypescript` would be truthy on any project that had TS entry points:

https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/src/index.js#L387

To allow projects with TS entry points to disable types from being output, we now simply disable TS's `declaration` compiler option if `options.generateTypes` is set to false.

**Does this PR introduce a breaking change?**

No

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -518,10 +518,12 @@ function createConfig(options, entry, format, writeMeta) {
 							tsconfigDefaults: {
 								compilerOptions: {
 									sourceMap: options.sourcemap,
-									declaration: true,
+									declaration: emitDeclaration,
 									allowJs: true,
 									emitDeclarationOnly: options.generateTypes && !useTypescript,
-									declarationDir: getDeclarationDir({ options, pkg }),
+									...(emitDeclaration && {
+										declarationDir: getDeclarationDir({ options, pkg }),
+									}),
 									jsx: 'preserve',
 									jsxFactory:
 										// TypeScript fails to resolve Fragments when jsxFactory


src/index.js
```diff
@@ -518,10 +518,12 @@ function createConfig(options, entry, format, writeMeta) {
...
 								compilerOptions: {
 									sourceMap: options.sourcemap,
-									declaration: true,
+									declaration: options.generateTypes !== false,
 									allowJs: true,
 									emitDeclarationOnly: options.generateTypes && !useTypescript,
-									declarationDir: getDeclarationDir({ options, pkg }),
+									...(options.generateTypes !== false && {
+										declarationDir: getDeclarationDir({ options, pkg }),
+									}),
```
rschristian, line 526: The behaviour here is the same as if we had simply set `declarationDir` to `undefined`, but rpt2 throws a bunch of errors  if that key exists at all & `declaration` is `false`.

```
$ microbundle --generateTypes false
rpt2: options error TS5069: Option 'declarationDir' cannot be specified without specifying option 'declaration' or option 'composite'.
rpt2: options error TS5069: Option 'declarationDir' cannot be specified without specifying option 'declaration' or option 'composite'.
rpt2: options error TS5069: Option 'declarationDir' cannot be specified without specifying option 'declaration' or option 'composite'.
rpt2: options error TS5069: Option 'declarationDir' cannot be specified without specifying option 'declaration' or option 'composite'.
```

test/__snapshots__/index.test.js.snap
```diff
@@ -2728,9 +2728,10 @@ ts-custom-declaration
     index.js.map
     index.umd.js
     index.umd.js.map
+  node_modules
```
developit, line 2731:
```suggestion
```

developit (approved):
LGTM! Just the one nit about `node_modules` in the snapshot

test/__snapshots__/index.test.js.snap
rschristian: Clicked "Commit suggestion" too fast there. What was not working w/ the CI? Seemed to be passing w/out any issues?

rpt2 creates some cache files in `node_modules`, so I believe it's needed.

Matches the other declaration test:

https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/test/__snapshots__/index.test.js.snap#L2775-L2795

PR #896, opened 2021-09-15
fix: Allows naming of outputs individually

### LINKED ISSUE #884: Bug: Shouldn't fall back to pkg name in the absence of output flag & pkg main
Packages that don't use the `output` flag or have a `"main"` in their `package.json` have their preferred filename output ignored.

For example, say I have the following package:

```
{
  "name": "foo",
  "exports": {
    ".": "./dist/bar.js",
    ...
  }
}
```

Unfortunately, because of the `replaceName()` here, the desired output is replaced by the package name instead.

https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/src/index.js#L297-L303

https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/src/index.js#L255-L260

`mainNoExtension` traces back to here, which defines a couple of fallbacks.

https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/src/index.js#L219-L225

I can't really say that I understand why this was done in the first place, so don't want to go "fixing" it as there might be some relying on this. To me, it seems like that in the absence of the `output` flag, these fields should be read and used exactly, no replacing.

`mainNoExtension` seems like it should be the last of the fallbacks, not something that overwrites valid entries & chosen names.

rschristian:
Closes #884

Each output can be individually named now, they won't all share the same name. This removes a reliance upon `"main"` to provide a custom name w/out the `--output` flag.


rschristian:
https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/src/index.js#L219-L225

I'd really love to remove `pkgMain` from the equation entirely here, though that would be a breaking change that I imagine would have a large effect on existing projects. Ideally, `"main"` should only affect whatever the main output is. If it's set to `./dist/foo-bar-baz.js`, that's fine, and in lieu of being also set, the other outputs should fall back to the package name. 

For example,

```json
{
  "name": "custom-output",
  "main": "./dist/index.js",
  "exports": "./dist/custom-name.js",
  "scripts": {
    "build": "microbundle -f cjs,es,umd,modern"
  }
}
```

should, in my opinion, generate the following output:

```
custom-output
  dist
    custom-name.js
    custom-name.js.map
    custom-output.esm.js
    custom-output.esm.js.map
    index.js
    index.js.map
    custom-output.umd.js
    custom-output.umd.js.map
```

However, right now, to avoid breaking changes, it generates the following:

```
custom-output
  dist
    custom-name.js
    custom-name.js.map
    index.esm.js
    index.esm.js.map
    index.js
    index.js.map
    index.umd.js
    index.umd.js.map
```

rschristian:
https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/src/index.js#L135

Reminder: Need to look at / edit this (`targetDir`) and probably the `--output` flag too. Can now output formats in different directories, if desired.

```json
{
  "main": "dist/index.js",
  "exports": "build/index.js"
}
```

^ actually works correctly now which probably throws a few behaviors for a loop.

src/index.js
```diff
@@ -138,10 +137,9 @@ export default async function microbundle(inputOptions) {
 		}),
 	);
 
-	const targetDir = relative(cwd, dirname(options.output)) || '.';
 	const sourceExist = options.input.length > 0;
 	const banner = sourceExist
-		? blue(`Build "${options.pkg.name}" to ${targetDir}:`)
+		? blue(`Built "${options.pkg.name}":`)
```
rschristian, line 142: `targetDir` is a bit unknowable now, so it's been removed.

src/index.js
```diff
@@ -279,42 +270,45 @@ function walk(exports, includeDefault) {
...
-			(pkg.syntax && pkg.syntax.esmodules) ||
-			pkg.esmodule ||
-			'x.modern.js',
-		mainNoExtension,
+	mainsByFormat.modern = resolve(
+		(!multipleEntries &&
+			((pkg.exports && walk(pkg.exports, pkgTypeModule)) ||
+				(pkg.syntax && pkg.syntax.esmodules) ||
+				pkg.esmodule)) ||
+			`${defaultOutputNoExtension}.modern.${pkgTypeModule ? 'js' : 'mjs'}`,
```
rschristian, line 303: Missed this in #802: if the CJS output can respond to package type, so should the modern output I believe.

src/index.js
```diff
@@ -279,42 +270,45 @@ function walk(exports, includeDefault) {
 function getMain({ options, entry, format }) {
 	const { pkg } = options;
 	const pkgMain = options['pkg-main'];
+	const pkgTypeModule = pkg.type === 'module';
+	let multipleEntries = options.multipleEntries;
```
rschristian, line 274: `multipleEntries` needs to act as an escape hatch to our name resolution logic to ensure no collisions.

This can be seen with Microbundle itself; when we create a `cli.js` and `microbundle.js`, we need to ensure the assigned name of `cli` does not get replaced by the `"main"` field ("microbundle").

src/index.js
```diff
@@ -279,42 +270,45 @@ function walk(exports, includeDefault) {
...
 	const mainsByFormat = {};
 
-	mainsByFormat.es = replaceName(
-		pkg.module && !pkg.module.match(/src\//)
-			? pkg.module
-			: pkg['jsnext:main'] || 'x.esm.js',
-		mainNoExtension,
+	mainsByFormat.es = resolve(
+		(!multipleEntries &&
+			(pkg.module && !pkg.module.match(/src\//)
```
rschristian, line 293: Random q: What's this match against `src/` about? Goes back to repo init afaict so couldn't find any context there.

PR #899, opened 2021-09-28
docs: Update obsolete `default` exports in package.json

donysukardi:
Current example uses `default`, which is no longer valid based the following code 

https://github.com/developit/microbundle/blob/b1a637486234a2ae784ccf0c512321e2d3efef7c/src/index.js#L262-L265

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ The filenames and paths for generated bundles in each format are defined by the 
   "module": "dist/foo.m.js",           // ES Modules output bundle 
   "exports": {
     "require": "./dist/foo.js",        // CommonJS output bundle
-    "default": "./dist/foo.modern.js", // Modern ES Modules output bundle
+    "import": "./dist/foo.modern.js", // Modern ES Modules output bundle
   },
   "types": "dist/foo.d.ts"             // TypeScript typings directory
 }


rschristian:
That's absolutely a valid key, just not one we check (at the moment) while generating output. Nothing obsolete there.

This could change at some point though, see https://github.com/developit/microbundle/pull/853#discussion_r670514651

donysukardi:
I see. I suppose it’s a regression then based on the linked issue. I copied the example in the readme and was surprised that it didn’t work

rschristian:
There's no regression, that was never supported.

donysukardi:
Then the example is incorrect given the current state

rschristian:
It's correct, just not fulfilling the behavior you wish.

sirianni:
I don't really understand the confusion here, clearly the example in the `README` is incorrect as @donysukardi points out.

image

rschristian:
The confusion is that they believed export keys were no longer valid based on us not checking for them. We don't check for every key of exports to generate output for, and never have. It'd certainly be nice to be able to, but that's not the point we're at.

The example is valid and correctly formed, no issues with it.

I believe there's a PR open for adding `default` to the list of keys to check against.

PR #912, opened 2021-11-25
Remove declaration dir warning when declarations are disabled

andyrichardson:
# About

Given the following tsconfig

```json
{
  "compilerOptions": {
    "declaration": false
  }
}
```

The user will see this message whenever building

> rpt2: options error TS5069: Option 'declarationDir' cannot be specified without specifying option 'declaration' or option 'composite'.

This PR introduces a check to only provide a `declarationDir` option to rollup if declarations aren't omitted.

# Other stuff

This is just a proposal - the `tsconfig.json` would need to be properly interpreted in order to get the `declaration` value (e.g. to cover `extends`).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -227,6 +227,14 @@ async function getOutput({ cwd, output, pkgMain, pkgName }) {
 }
 
 function getDeclarationDir({ options, pkg }) {
+	try {
+		const tsconfig = JSON.parse(fs.readFileSync(options.tsconfig || "tsconfig.json", 'utf-8'))
+
+		if (tsconfig && tsconfig.compilerOptions && tsconfig.compilerOptions.declaration === false) {
+			return undefined;
+		}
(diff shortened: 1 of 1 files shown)


rschristian:
I'm thinking we can just alter [`emitDeclaration`](https://github.com/developit/microbundle/blob/778ca3ce3e75370ca0f4802b224891bda2417704/src/index.js#L393-L396) to check for `tsconfig.compilerOptions.declaration` too (and accordingly change [the condition for adding `declarationDir` to `emitDeclaration`](https://github.com/developit/microbundle/blob/778ca3ce3e75370ca0f4802b224891bda2417704/src/index.js#L536))

Check `options.generateTypes` then `tsconfig.compilerOptions.declaration` and finally `!!(pkg.types || pkg.typings)`. That should cover all the bases

PR #914, opened 2021-12-09
Rollup tweaks for output size/perf

developit:
This turns off `externalLiveBindings`, which generates getters for re-exported bindings.
It also enables modern code generation for the modern bundles (arrow functions, object shorthand, etc).


probablyup:
@developit this change sounds good - picking it back up eventually?

developit:
Yes - I think we are on more recent Rollup now too and can enable es2015 output.

PR #919, opened 2021-12-22
fix: Corrects CSS output being overwritten when .cjs ext used

### LINKED ISSUE #883: .cjs extension in `main` package.json field causes imported CSS file output to be overwritten
Using a `.cjs` extension in the `main` field of `package.json` (e.g. `"main": "./dist/foo.cjs"`) as per README causes microbundle to swallow any .css or .scss files imported

## Reproduction steps

Repo: https://github.com/lokimckay-references/microbundle-883

1. `npm init -y`
1. `npm install microbundle --save-dev`
1. add `styles.css` and `index.js` files
1. `import "./styles.css"` within `index.js`
1. copy source/exports/main/module/unpkg fields into `package.json` as per README
    ```
    {
      "name": "foo",                     // your package name
      "type": "module",
      "source": "src/foo.js",            // your source code
      "exports": "./dist/foo.modern.js", // where to generate the modern bundle (see below)
      "main": "./dist/foo.cjs",          // where to generate the CommonJS bundle
      "module": "./dist/foo.module.js",  // where to generate the ESM bundle
      "unpkg": "./dist/foo.umd.js",      // where to generate the UMD bundle (also aliased as "umd:main")
      "scripts": {
        "build": "microbundle",          // compiles "source" to "main"/"module"/"unpkg"
        "dev": "microbundle watch"       // re-build when source files change
      }
    }
    ```
1. update `source` field to point to `index.js`
1. `npm run build`

## Actual output

- All compiled formats of index.js
- Below error message

```
The emitted file "foo.cjs" overwrites a previously emitted file of the same name.
```

image

## Expected output

- All compiled formats of index.js
- **styles.css**
- No error message

## Workaround

Change the file extension of the `main` field from `./dist/foo.cjs` -> `./dist/foo.js`
= achieves expected output

rschristian:
Fixes #883

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,14 @@ import typescript from 'rollup-plugin-typescript2';
 import json from '@rollup/plugin-json';
 import OMT from '@surma/rollup-plugin-off-main-thread';
 import logError from './log-error';
-import { isDir, isFile, stdout, isTruthy, removeScope } from './utils';
+import {
+	EXTENSION,
+	isDir,
+	isFile,
+	isTruthy,
+	stdout,
+	removeScope,
+} from './utils';
 import { getSizeInfo } from './lib/compressed-size';
 import { normalizeMinifyOptions } from './lib/terser';
 import {
@@ -279,17 +286,12 @@ function getMain({ options, entry, format }) {
 
 	let mainNoExtension = options.output;
 	if (options.multipleEntries) {
-		let name = entry.match(
-			/([\\/])index(\.(umd|cjs|es|m))?\.(mjs|cjs|[tj]sx?)$/,
-		)
+		let name = entry.match(new RegExp(/([\\/])index/.source + EXTENSION.source))
 			? mainNoExtension
 			: entry;
 		mainNoExtension = resolve(dirname(mainNoExtension), basename(name));
 	}
-	mainNoExtension = mainNoExtension.replace(
-		/(\.(umd|cjs|es|m))?\.(mjs|cjs|[tj]sx?)$/,
-		'',
-	);
+	mainNoExtension = mainNoExtension.replace(EXTENSION, '');
 
 	const mainsByFormat = {};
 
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -483,7 +481,7 @@ function createConfig(options, entry, format, writeMeta) {
 							!!writeMeta &&
 							options.css !== 'inline' &&
 							options.output.replace(
-								/(\.(umd|cjs|es|m))?\.(mjs|[tj]sx?)$/,
+								/(\.(umd|cjs|es|m))?\.([cm]js|[tj]sx?)$/,
```
rschristian, line 484: This was the only regex that needed to change, the others were functionally fine, but I updated them to match.

MinJieLiu:
When can this pr be merged

src/utils.js
```diff
@@ -43,3 +43,5 @@ export function safeVariableName(name) {
 	const identifier = normalized.replace(INVALID_ES3_IDENT, '');
 	return camelCase(identifier);
 }
+
+export const EXTENSION = /(\.(umd|cjs|es|m))?\.([cm]?[tj]sx?)$/;
```
rschristian, line 47: Names are hard... 😅

Certainly open to suggestions

src/index.js
```diff
@@ -482,10 +484,7 @@ function createConfig(options, entry, format, writeMeta) {
 						extract:
 							!!writeMeta &&
 							options.css !== 'inline' &&
-							options.output.replace(
-								/(\.(umd|cjs|es|m))?\.(mjs|[tj]sx?)$/,
-								'.css',
-							),
+							options.output.replace(EXTENSION, '.css'),
```
rschristian, line 487: This is the only regex that actually needed to be corrected. #802 missed this when adding in support, which is totally understandable given that the regex existed in 3 separate locations. Hopefully extracting this out makes it easier to handle.

src/utils.js
rschristian: Also (horrifyingly) supports the [`.cts` and `.mts` extensions](https://www.typescriptlang.org/docs/handbook/esm-node.html), as it looks like TS might head down that path

PR #882, opened 2021-08-22
feat: :sparkles: Closes #497: preserve trailing newline in mangle.json

MiKr13:
Fixes #497 

This commit adds editorconfig & index.js changes to enable trailing new line detection & addition in mangle.json

PR #926, opened 2022-01-20
Fix mangle.json being overwritten with [object Object]

developit:
Regression from #882.

### DIFF
--- a/.changeset/gorgeous-singers-invite.md
+++ b/.changeset/gorgeous-singers-invite.md
@@ -1,0 +1,5 @@
+---
+"microbundle": patch
+---
+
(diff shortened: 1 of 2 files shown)


rschristian (approved):
Huh, wonder what's tripping up the changset-bot comment? I noticed it the other night in [wmr too](https://github.com/preactjs/wmr/pull/901#issuecomment-1016040467). Seems to not update itself anymore.

developit:
Funky, yeah. I just peeked and there are no issues mentioning this on changesets/changesets or changesets/bot. The logic for detecting added changesets seems pretty straightforward. Maybe I clicked too fast? haha

Issue #934, opened 2022-03-09
NPE occured when typescript plugin goes to wrong


mycoin:
version: 0.13.3

error details:
```
(rpt2 plugin) Error: /Users/apple/Workspace/Odts/starter-sdk/src/index.ts(13,16): semantic error TS2552: Cannot find name 'StarterSdk2s'. Did you mean 'StarterSdk'?
Error: /Users/apple/Workspace/Odts/starter-sdk/src/index.ts(13,16): semantic error TS2552: Cannot find name 'StarterSdk2s'. Did you mean 'StarterSdk'?
    at error (/Users/apple/Zend/bundle-scripts/node_modules/rollup/dist/shared/rollup.js:161:30)
    at throwPluginError (/Users/apple/Zend/bundle-scripts/node_modules/rollup/dist/shared/rollup.js:21832:12)
    at Object.error (/Users/apple/Zend/bundle-scripts/node_modules/rollup/dist/shared/rollup.js:22557:20)
    at Object.error (/Users/apple/Zend/bundle-scripts/node_modules/rollup/dist/shared/rollup.js:22009:42)
    at RollupContext.error (/Users/apple/Zend/bundle-scripts/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:17237:30)
    at /Users/apple/Zend/bundle-scripts/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:25033:23
    at arrayEach (/Users/apple/Zend/bundle-scripts/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:545:11)
    at Function.forEach (/Users/apple/Zend/bundle-scripts/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:9397:14)
    at printDiagnostics (/Users/apple/Zend/bundle-scripts/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:25006:12)
    at Object.transform (/Users/apple/Zend/bundle-scripts/node_modules/rollup-plugin-typescript2/dist/rollup-plugin-typescript2.cjs.js:29275:17)

TypeError: Cannot read property 'then' of undefined
    at WatchEmitter.<anonymous> (/Users/apple/Zend/bundle-scripts/node_modules/microbundle/dist/microbundle.js:720:29)
    at WatchEmitter.emit (events.js:314:20)
[...]

PR #935, opened 2022-03-09
feat: [auto-commit] fix: _sizeInfo npe

mycoin:
fix NPE occured when typescript plugin goes to wrong

#934

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -177,7 +177,6 @@ function doWatch(options, cwd, steps) {
 					}
 				}
 				if (e.code === 'END') {
-					// fix NPE
 					if (options._sizeInfo) {
 						options._sizeInfo.then(text => {
 							stdout(`Wrote ${text.trim()}`);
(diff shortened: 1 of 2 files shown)


ForsakenHarmony (approved):

src/index.js
```diff
@@ -177,9 +177,12 @@ function doWatch(options, cwd, steps) {
...
 				if (e.code === 'END') {
-					options._sizeInfo.then(text => {
-						stdout(`Wrote ${text.trim()}`);
-					});
+					// fix NPE
+					if (options._sizeInfo) {
+						options._sizeInfo.then(text => {
+							stdout(`Wrote ${text.trim()}`);
+						});
+					}
```
rschristian, line 185:
```suggestion
					if (options._sizeInfo) {
						options._sizeInfo.then(text => {
							stdout(`Wrote ${text.trim()}`);
						});
					}
```

nit: An undefined check is obvious enough to not require a comment.

developit (approved):

PR #941, opened 2022-03-25
fix: tsconfig cwd differing from microbundle's cwd

### LINKED ISSUE #939: `--cwd` and `--tsconfig`
Is it expected that `tsconfig` resolves against the calling directory even when `--cwd` is specified? I have a use case where I have a browser lib and a node lib in the same repo that have distinct `tsconfig.json` files. Currently I have to specify the full path to make `tsconfig.json` discovery work, that is `--cwd a/b/c --tsconfig a/b/c/tsconfig.json` instead of just `--cwd a/b/c --tsconfig tsconfig.json` or just `--cwd a/b/c`. Is there some other configuration that I do not know about that would make this more intuitive?

rschristian:
Closes #939

Simply ensures the TS plugin will begin its search for a `tsconfig.json` in the specified `--cwd`.

This is not a breaking change; in the case of a `tsconfig.json` only existing higher in a project structure, and not in the passed `--cwd`, rpt2 (which makes use of `typescript` itself, actually) will ascend the file tree attempting to discover a config file.

### DIFF
--- a/.changeset/clever-chicken-end.md
+++ b/.changeset/clever-chicken-end.md
@@ -1,0 +1,5 @@
+---
+'microbundle': patch
+---
+
+Ensures TS plugin will begin its search for a 'tsconfig.json' in the set cwd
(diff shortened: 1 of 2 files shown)


developit (approved):
Awesome.

PR #947, opened 2022-04-20
fix: babel exclude for hoisted modules

### LINKED ISSUE #946: Inlined dependency shaken different when symlinked
## Summary

Bit of a weird one, but I'm seeing different results from Microbundle based on whether a dependency is symlinked or not. The module is resolved without issue in both cases.

As far as I can tell this is not specific to any one dependency.

## Reproduction

1. Clone https://github.com/rschristian/microbundle-symlink-bug
2. `$ yarn install && yarn example build`
3. Note that the bundle is 4.46 kb w/ brotli
4. `$ rm -rf **/node_modules`
5. Edit the root `package.json` to look like the following (disables hoisting):
```json
{
  "name": "twind-microbundle",
  "private": true,
  "scripts": {
    "example": "yarn workspace example"
  },
  "workspaces": {
    "packages": ["packages/*"],
    "nohoist": ["**/example/**"]
  }
}
```
7. `$ yarn install && yarn build`
8. Note that the bundle is now 1.34 kb w/ brotli

rschristian:
Closes #946

### DIFF
--- a/.changeset/hip-olives-allow.md
+++ b/.changeset/hip-olives-allow.md
@@ -1,0 +1,5 @@
+---
+'microbundle': patch
+---
+
+Ensures hoisted node_modules are excluded from babel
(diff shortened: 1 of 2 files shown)


developit (approved):

developit:
nice! I didn't realize the string exclude there was anchored.

PR #950, opened 2022-04-25
refactor: Output ESM files w/ `.mjs` when appropriate

rschristian:
Per the conversation on Slack, both modern and esm output formats may now create `.mjs` files when appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ export default function (e, t, r) {
 ```jsonc
 {
 	"main": "./dist/foo.umd.js", // legacy UMD output (for Node & CDN use)
-	"module": "./dist/foo.module.js", // legacy ES Modules output (for bundlers)
+	"module": "./dist/foo.module.mjs", // legacy ES Modules output (for bundlers)
 	"exports": "./dist/foo.modern.mjs", // modern ES2017 output
 	"scripts": {
 		"build": "microbundle src/foo.js"
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ export default function (e, t, r) {
 {
 	"main": "./dist/foo.umd.js", // legacy UMD output (for Node & CDN use)
 	"module": "./dist/foo.module.js", // legacy ES Modules output (for bundlers)
-	"exports": "./dist/foo.modern.js", // modern ES2017 output
+	"exports": "./dist/foo.modern.mjs", // modern ES2017 output
 	"scripts": {
 		"build": "microbundle src/foo.js"
 	}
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -121,7 +121,7 @@ export default function (e, t, r) {
 {
 	"main": "./dist/foo.umd.js", // legacy UMD output (for Node & CDN use)
 	"module": "./dist/foo.module.js", // legacy ES Modules output (for bundlers)
```
rschristian, line 123:
```suggestion
	"module": "./dist/foo.module.mjs", // legacy ES Modules output (for bundlers)
```

developit: whoops, good catch.

src/index.js
```diff
@@ -300,18 +301,22 @@ function getMain({ options, entry, format }) {
 	mainsByFormat.es = replaceName(
 		pkg.module && !pkg.module.match(/src\//)
 			? pkg.module
-			: pkg['jsnext:main'] || 'x.esm.js',
+			: pkg['jsnext:main'] || pkgTypeModule
+			? 'x.esm.js'
+			: 'x.esm.mjs',
```
developit, line 306: I think `.m.js` and `.esm.js` were both poor approximations of `.mjs`, and we should be able to drop the infix.

```suggestion
			: 'x.mjs',
```

README.md
rschristian: Nothing like an approval to make mistakes stand out 😅

src/index.js
rschristian: Makes sense, though I wonder if that will confuse some users; for users who might not be super familiar with module formats, Microbundle is quite easy to understand given the current names. `esm` -> `x.esm.js`, `umd` -> `x.umd.js`, etc.

src/index.js
```diff
@@ -439,6 +445,15 @@ function createConfig(options, entry, format, writeMeta) {
...
 
+	// Warn about the (somewhat) breaking change in #950
+	if (format === 'es' && !pkg.module && outputEntryFileName.endsWith('.mjs')) {
+		stdout(
+			yellow(
+				'Warning: If "module" is not specified in your package.json, Microbundle will now output the ESM format with ".mjs" as the file extension in CJS packages. If this is not the desired behavior, please specify "module".',
+			),
+		);
+	}
+
```
rschristian, line 456: New warning message as discussed

src/index.js
developit: It's a reasonable thought, yeah. I could go either way on this.

src/index.js
```diff
@@ -439,6 +445,15 @@ function createConfig(options, entry, format, writeMeta) {
 	const outputDir = dirname(absMain);
 	const outputEntryFileName = basename(absMain);
 
+	// Warn about the (somewhat) breaking change in #950
+	if (format === 'es' && !pkg.module && outputEntryFileName.endsWith('.mjs')) {
+		stdout(
+			yellow(
+				'Warning: If "module" is not specified in your package.json, Microbundle will now output the ESM format with ".mjs" as the file extension in CJS packages. If this is not the desired behavior, please specify "module".',
```
developit, line 452:
```suggestion
				'Warning: your package.json does not specify {"type":"module"}. Microbundle assumes this is a CommonJS package and is generating ES Modules with the ".mjs" file extension.',
```

PR #952, opened 2022-04-28
Remove git.io usages

mhmdanas:
See https://github.blog/changelog/2022-04-25-git-io-deprecation/.

Apologies in advance for the incoming barrage of PRs doing the exact same thing, hope you don't mind.


developit (approved):
all good! thanks for the service :)

PR #956, opened 2022-05-04
chore: Silence warnings when using 'node:' protocol

rschristian:
The current solution for matching builtins doesn't catch imports using the new-ish `node:` protocol:

```
Failed to resolve the module node:http imported by src/index.js
Is the module installed? Note:
 ↳ to inline a module into your bundle, install it to "devDependencies".
 ↳ to depend on a module via import/require, install it to "dependencies".
```

This just adds it to the patterns to match against when the target for Microbundle is Node.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -347,7 +347,7 @@ function createConfig(options, entry, format, writeMeta) {
 	// We want to silence rollup warnings for node builtins as we rollup-node-resolve threats them as externals anyway
 	// @see https://github.com/rollup/plugins/tree/master/packages/node-resolve/#resolving-built-ins-like-fs
 	if (options.target === 'node') {
-		external = external.concat(builtinModules);
+		external = [/node:.*/].concat(builtinModules);
 	}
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -347,7 +347,7 @@ function createConfig(options, entry, format, writeMeta) {
 	// We want to silence rollup warnings for node builtins as we rollup-node-resolve threats them as externals anyway
 	// @see https://github.com/rollup/plugins/tree/master/packages/node-resolve/#resolving-built-ins-like-fs
 	if (options.target === 'node') {
-		external = external.concat(builtinModules);
+		external = [/node:.*/].concat(builtinModules);
```
rschristian, line 350: I purposefully replaced `external` here as otherwise we get duplicates.

https://github.com/developit/microbundle/blob/6c59174aadefb5ac5cf5e7670a750c4175b1c0d5/src/index.js#L339

Unless this was intentional as slightly faster path for those 4 builtins specifically?

developit (approved):

PR #960, opened 2022-05-16
Fix problem

zyrong:
fix: css compilation problems when multiple entries

fix: Fix problems related to use rollup-plugin-terser
1. The nameCache obtained in writebundle hook is not a reference to terseroptions
2. JSON.stringify(nameCache) requires converting regex to a string first.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -111,12 +111,7 @@ export default async function microbundle(inputOptions) {
 	for (let i = 0; i < options.entries.length; i++) {
 		for (let j = 0; j < formats.length; j++) {
 			steps.push(
-				createConfig(
-					options,
-					options.entries[i],
-					formats[j],
-					i === 0 && j === 0,
-				),
+				createConfig(options, options.entries[i], formats[j], j === 0),
 			);
 		}
 	}
@@ -333,6 +328,7 @@ function getMain({ options, entry, format }) {
 const shebang = {};
 
 function createConfig(options, entry, format, writeMeta) {
+	const entryFileName = basename(entry, extname(entry));
 	let { pkg } = options;
 
 	/** @type {(string|RegExp)[]} */
@@ -502,7 +498,7 @@ function createConfig(options, entry, format, writeMeta) {
 						extract:
 							!!writeMeta &&
 							options.css !== 'inline' &&
-							options.output.replace(EXTENSION, '.css'),
+							resolve(outputDir, `${entryFileName}.css`),
 						minimize: options.compress,
 						sourceMap: options.sourcemap && options.css !== 'inline',
 					}),
--- a/src/index.js
+++ b/src/index.js
@@ -386,15 +386,15 @@ function createConfig(options, entry, format, writeMeta) {
 
 	// let rollupName = safeVariableName(basename(entry).replace(/\.js$/, ''));
 
-	let nameCache = {};
-	const bareNameCache = nameCache;
 	// Support "minify" field and legacy "mangle" field via package.json:
 	const rawMinifyValue = options.pkg.minify || options.pkg.mangle || {};
-	let minifyOptions = typeof rawMinifyValue === 'string' ? {} : rawMinifyValue;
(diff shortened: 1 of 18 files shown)


rschristian:
Would you mind splitting half of this out into another PR? Fixing two (seemingly) different issues in one PR makes it a bit hard to review and follow.

For both, do you mind writing up issues showcasing the problems you're experiencing?

> css compilation problems when multiple entries

This is a bit vague, for example. It's difficult to review the work because I don't know what issues you've run into or if your added test cases adequately check your fix.

Thanks

zyrong:
> Would you mind splitting half of this out into another PR? Fixing two (seemingly) different issues in one PR makes it a bit hard to review and follow.
> 
> Edit: Also helps the history. With both fixes in one PR, it's a bit murky which bit of code fixes what issue. If there's ever problems, it helps if it's clear to what issue a line was altered/added.
> 
> For both, do you mind writing up issues showcasing the problems you're experiencing?
> 
> > css compilation problems when multiple entries
> 
> This is a bit vague, for example. It's difficult to review the work because I don't know what issues you've run into or if your added test cases adequately check your fix.
> 
> Thanks

What should I do to split into two pr

1. This commit fixes that when different entries reference different CSS, only the related CSS of the first entry will be packaged. For specific changes, please see this commit. I also added the corresponding test case 
[fix: css compilation problems when multiple entries](https://github.com/developit/microbundle/pull/960/commits/4580e5fac6879aecf7fd62d63dcb3687406fc50d)

2. This commit fixes the problems related to terser namecache. When terser compresses the code, the damaged variable names will be written to the namecache. Finally, the namecache will be written to the file system in [this code](https://github.com/developit/microbundle/blob/9a4e2b2096d3824661738edb142b4658cf3d9d0b/src/index.js#L640), but after my test, it doesn't seem to work here
[fix: Fix problems related to use rollup-plugin-terser](https://github.com/developit/microbundle/pull/960/commits/b0f22912ad4e6f215f272202f528311a38643be8)

rschristian:
> What should I do to split into two pr

CSS fixes in one, Terser thing in another. Specific commits get lost as PRs are squashed and merged.

src/index.js
```diff
@@ -333,6 +328,7 @@ function getMain({ options, entry, format }) {
 const shebang = {};
 
 function createConfig(options, entry, format, writeMeta) {
+	const entryFileName = basename(entry, extname(entry));
```
rschristian, line 331: This is incorrect. Please use the existing naming setup that we had, it ensures names are properly stripped:

https://github.com/developit/microbundle/blob/9a4e2b2096d3824661738edb142b4658cf3d9d0b/src/index.js#L505

`index.esm.js` as an input needs to become `index`; yours would have it become `index.esm`, so `index.esm.css` would be output. We need to strip indicators towards ES module format, i.e., `cjs`, `es[m]`, etc.

src/index.js
```diff
@@ -502,7 +475,7 @@ function createConfig(options, entry, format, writeMeta) {
 						extract:
 							!!writeMeta &&
 							options.css !== 'inline' &&
-							options.output.replace(EXTENSION, '.css'),
+							resolve(outputDir, `${entryFileName}.css`),
```
rschristian, line 478: I'd rather we not use `outputDir` as it isn't guaranteed to match up with expectations in the future (see #896)

CSS should continue to be output according to the result of [`getOutput`](https://github.com/developit/microbundle/blob/9a4e2b2096d3824661738edb142b4658cf3d9d0b/src/index.js#L231-L237) for resilience IMO.

zyrong: How about changing it to [this](https://github.com/developit/microbundle/pull/961)

zyrong:
> > What should I do to split into two pr
> 
> CSS fixes in one, Terser thing in another. Specific commits get lost as PRs are squashed and merged.

I have split it into two independent PRs and modified them. Can you make some suggestions or reply to the changes

PR #961, opened 2022-05-17
fix: css compilation problems when multiple entries

zyrong:
fix: css compilation problems when multiple entries

This commit fixes that when different entries reference different CSS, only the related CSS of the first entry will be packaged. For specific changes, please see this commit. I also added the corresponding test case
[Original PR](https://github.com/developit/microbundle/pull/960)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -497,7 +497,7 @@ function createConfig(options, entry, format, writeMeta) {
 						extract:
 							!!writeMeta &&
 							options.css !== 'inline' &&
-							absMain.replace(EXTENSION, `.css`),
+							absMain.replace(EXTENSION, '.css'),
 						minimize: options.compress,
 						sourceMap: options.sourcemap && options.css !== 'inline',
 					}),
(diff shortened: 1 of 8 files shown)


src/index.js
```diff
@@ -502,7 +497,12 @@ function createConfig(options, entry, format, writeMeta) {
 						extract:
 							!!writeMeta &&
 							options.css !== 'inline' &&
-							options.output.replace(EXTENSION, '.css'),
+							(options.multipleEntries
+								? resolve(
+										dirname(options.output),
+										`${basename(entry.replace(EXTENSION, ''))}.css`,
+								  )
+								: options.output.replace(EXTENSION, `.css`)),
```
rschristian, line 505:
```suggestion
							absMain.replace(EXTENSION, '.css'),
```

We should be able to use `absMain` here instead to vastly simplify things. Should've been using it (rather than the `options.output.replace...`) from the start, but that's my fault.

rschristian (commented):
Sorry for the long delay after commenting on the other PR, life just got busy.

src/index.js
zyrong: 👌，I have already made changes to this

rschristian (approved):

PR #962, opened 2022-05-18
Remove mention of UMD support for web workers

nickbabcock:
OMT only supports amd or esm as output, so the readme should not
mention umd as supported.

When executing microbundle with umd format and `--workers`, the web
worker will fail to be detected or bundled.


rschristian (approved):
Good spot, thanks for correcting this!

PR #966, opened 2022-05-20
fix: Fix problems related to use terser nameCache

zyrong:
1. The nameCache obtained in writebundle hook is not a reference to terseroptions
2. JSON.stringify(nameCache) requires converting regex to a string first.

### DIFF
--- a/test/__snapshots__/index.test.js.snap
+++ b/test/__snapshots__/index.test.js.snap
@@ -701,6 +701,17 @@ exports[`fixtures build basic-multi-source with microbundle 8`] = `
 "
 `;
 
+exports[`fixtures build basic-multi-source-css with microbundle 1`] = `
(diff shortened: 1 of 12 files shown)


test/__snapshots__/index.test.js.snap
```diff
@@ -701,6 +701,17 @@ exports[`fixtures build basic-multi-source with microbundle 8`] = `
...
+"Used script: microbundle
+
+Directory tree:
+
+basic-multi-source-css
+
+
+Error: No entry module found for \\"basic-multi-source-css\\""
+`;
+
```
rschristian, line 714: Looks like you didn't remove the empty directory that Git left after switching branches
```suggestion
```

zyrong: yes

PR #968, opened 2022-06-03
fix: Incorrect transform with async/await (#565)

### LINKED ISSUE #565: Incorrect transform with async/await
microbundle uses the babel plugin `babel-plugin-transform-async-to-promises` which has this [serious bug](https://github.com/rpetrich/babel-plugin-transform-async-to-promises/issues/49) as it changes the behavior of your code once transpilled.  
Sharing that here as this could save a few hours of headache and insanity to someone. I wasn't that lucky.

PeterBurner:
* Update babel-plugin-transform-async-to-promises to v0.8.18
* fix #565


rschristian (approved):
Thanks!

Any idea if this is necessary? Looks like Yarn (v1, anyhow) will correctly resolve the newer version on a fresh install. Not sure if other tools do the same.

Edit: Looks like the snapshots need to be updated. Looks to be pretty minor & innocent

PeterBurner:
> Any idea if this is necessary? Looks like Yarn (v1, anyhow) will correctly resolve the newer version on a fresh install. Not sure if other tools do the same.

This might be true. However since the fixed bug in `babel-plugin-transform-async-to-promises` affects how code is transformed, an update might break things for some users. In my eyes this warrants a new microbundle version. Including a explicit warning in the release notes.

rschristian:
> an update might break things for some users

If it was breaking (I don't think it is), this should've been marked as a semver minor (as we're still in a 0-major), not a patch. But as `microbundle` used `^0.8.15`, lock file permitting, this update could've happened at any point. It's not locked behind a new version.

Fair enough if that was the justification, a new `microbundle` version will force users to get this fix I suppose.

PR #975, opened 2022-07-17
PoC: SWC

### LINKED ISSUE #943: Option to use a faster transpile/bundle engine?
esbuild and swc come to mind, being able to sidestep TSC for build speed would be amazing.

https://www.npmjs.com/package/rollup-plugin-swc
https://www.npmjs.com/package/rollup-plugin-esbuild

mmkal:
Hi! Was going to write this as a comment but I thought a functional PR would be a nicer way to make a proposal.

A cleaned up version of this could resolve #943 in a way that didn't involve either inflating microbundle's size or completely rewriting the whole library around SWC.

Basically, it looks for the `@swc/core` dependency, and if it's found, assumes that the user wants to use `swc` instead of babel or typescript. But `@swc/core` _isn't_ a dependency of microbundle. Which means most users can just continue using microbundle as is and not be affected. People who want to try microbundle out with SWC just have to install `@swc/core` which has the nice properties of being natural and intuitive for people who do want to use SWC, and unusual for anyone else.

It _does_ add `rollup-plugin-swc3` as a proper dependency of microbundle, because that's just a small javascript module, and it seems like the sort of thing microbundle should save users from having to worry about (whether they should use `rollup-plugin-swc`, `rollup-plugin-swc2` or `rollup-plugin-swc3`, and how to configure the chosen package).

I tried on a small typescript project with no dependencies, with an output of about 4kb. With swc it took about 6s, without it took 21s.

Some things that should be addressed before this PoC were to be shippable (I didn't attempt to do this because I am not very knowledgeable about microbundle, babel, or SWC):

1. there should be a CLI flag to let people who have `@swc/core` installed for unrelated reasons to opt out (e.g. `--swc false`)
2. it should be configured properly to output the right style in various configurations. In this PR I just hardcoded `target: 'es2018'` fairly arbitrarily. It should also respect any existing swc config detected, if present.
3. docs!
4. the rollup-plugin-swc3 minifier should be used instead of rollup-plugin-terser - probably a bunch of the perf benefits are being negated by not doing this
5. (maybe) rely on rollup-plugin-swc3 to emit typescript declaration files too. Might or might not be faster 🤷 
6. prettier/code cleanup etc. I deliberately avoided formatting `src/index.js` to make the diff easy to ready, but probably some refactoring should be done

If this goes well and people start using it, after a few months of the community helping flush out any issues, maybe something like the solution suggested by @developit in https://github.com/developit/microbundle/issues/943#issuecomment-1146963352 could be done with SWC. But the advantage of doing this way first is that we could likely do _exactly_ the same thing with [rollup-plugin-esbuild](https://www.npmjs.com/package/rollup-plugin-esbuild), and effectively support users with both. The community seems pretty evenly split between the two, and they're both an order of magnitude faster than typescript or babel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
 		"rollup": "^2.35.1",
 		"rollup-plugin-bundle-size": "^1.0.3",
 		"rollup-plugin-postcss": "^4.0.0",
+		"rollup-plugin-swc3": "^0.3.0",
 		"rollup-plugin-terser": "^7.0.2",
 		"rollup-plugin-typescript2": "^0.32.0",
 		"rollup-plugin-visualizer": "^5.6.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,6 +43,7 @@
 				"rollup": "^2.35.1",
 				"rollup-plugin-bundle-size": "^1.0.3",
 				"rollup-plugin-postcss": "^4.0.0",
+				"rollup-plugin-swc3": "^0.3.0",
 				"rollup-plugin-terser": "^7.0.2",
 				"rollup-plugin-typescript2": "^0.32.0",
 				"rollup-plugin-visualizer": "^5.6.0",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -107,6 +107,7 @@
 		"rollup": "^2.35.1",
 		"rollup-plugin-bundle-size": "^1.0.3",
 		"rollup-plugin-postcss": "^4.0.0",
+		"rollup-plugin-swc3": "^0.3.0",
```
probablyup, line 110: You can add this config to `package.json` for peer deps so we pin to a particular major version of swc but keep it optional:
```json
"peerDependenciesMeta": {
  "@swc/core": {
    "optional": true
  },
}
```

probablyup:
Thanks so much for taking a first stab at this Misha!

probablyup:
We'll want to look into a way to configure SWC plugins somehow without directly involving microbundle. Does SWC offer some sort of `.swcrc` sort of config mechanism it'll just pick up when running over a codebase?

rschristian:
Looks like there is a `.swcrc`, yes: https://swc.rs/docs/usage/core#options

probablyup:
@developit curious to get your thoughts on this!

rschristian:
Jason's on vacation right now so a response/review might be a bit.

> 3. (maybe) rely on rollup-plugin-swc3 to emit typescript declaration files too. Might or might not be faster 🤷

@mmkal Have you used it for this purpose in the past? Any major issues/shortcomings you know of?

We use `rollup-plugin-typescript2` which admittedly there are many problems with, so I'm curious if swc might be viable.

PR #979, opened 2022-07-24
upgrade all deps

JoviDeCroock:
This upgrades all dependencies, things to look into:

- new export default format reason why
- adding `@babel/plugin-proposal-private-property-in-object` silences loose warning but we might need to omit it for custom babelrc's


JoviDeCroock:
Closing due to inactivity

PR #988, opened 2022-08-10
fix: Strips jsxFactory from tsconfig if --imporSource is set

### LINKED ISSUE #857: react-jsx does not work with Preact
Hi

## Error
I've recently made the switch to microbundle for a project that uses Preact and TypeScript.
Following the docs, and me having the latest version of TypeScript, I put the following in my tsconfig:

```json
{
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "preact",
    //...
  }
}
```

However when building with microbundle I get the following error:
`rpt2: options error TS5089: Option 'jsxFactory' cannot be specified when option 'jsx' is 'react-jsx'.`
This seems weird behaviour as I have not specified the jsxFactory.

Combined with this I also get an error about the 'jsxFragmentFactory' having to be provided to use JSX Fragments, but I presume that is due to the fact the the jsxFactory is somehow incorrect.

I tried hardcoding the values from the config in my script (without success) by doing the following

`microbundle --jsx react-jsx --jsxImportSource preact`

## Possible solution

I've been able to get microbundle working if I switch back to the tsconfig for older versions of TypeScript, so I'll be opting into this behaviour for now. For reference this is what the old config is:

```json
{
  "compilerOptions": {
    "jsx": "react-jsx",
    "jsxImportSource": "preact",
    //...
  }
}
```

rschristian:
Closes #857

If a user supplies `--jsxImportSource` we should remove `jsxFactory` and `jsxFragmentFactory` from the default `tsconfig` as they're mutually exclusive with `jsx: 'react-jsx'`.


src/index.js
```diff
@@ -549,9 +550,15 @@ function createConfig(options, entry, format, writeMeta) {
...
-									jsxFragmentFactory: options.jsxFragment,
+									...(!options.jsxImportSource && {
+										jsx: 'preserve',
+										jsxFactory: options.jsx,
+										jsxFragmentFactory: options.jsxFragment,
+									}),
+									...(options.jsxImportSource && {
+										jsx: 'react-jsx',
+										jsxImportSource: options.jsxImportSource,
+									}),
```
rschristian, line 561: Theoretically we should be fine to leave `jsx` untouched & remove that second conditional properties block, but that relies on the assumption that a user has a `tsconfig.json` to use.

Conditionally adding `jsxFactory` and `jsxFragmentFactory` is enough to fix the errors, but somehow `jsx` will need to become `'react-jsx'` and `jsxImportSource` will need to be set to result in valid typings. While users will likely have a correct `tsconfig.json` that will get merged over our defaults, if they don't, the types will just contain a bunch of `any`.

Setting these ourselves allows us to make fewer assumptions about the user's configs that are external to microbundle.

src/index.js
```diff
@@ -549,9 +550,16 @@ function createConfig(options, entry, format, writeMeta) {
...
+									...(options.jsxImportSource
+										? {
+												jsx: 'react-jsx',
+												jsxImportSource: options.jsxImportSource,
+										  }
+										: {
+												jsx: 'preserve',
+												jsxFactory: options.jsx,
+												jsxFragmentFactory: options.jsxFragment,
+										  }),
```
rschristian, line 562: Theoretically we would be fine to leave `jsx` untouched & remove that first piece of the ternary, however, that relies on the assumption that the user does have a `tsconfig.json` and that it's correctly set.

Conditionally adding `jsxFactory` (and `jsxFragmentFactory`) is enough to quell the errors we were getting in the linked issue, but when using `--jsxImportSource` with TS, the `tsconfig.json` will need to have different values. While users in all likelihood will have a `tsconfig.json` that's correctly configured in that case, if they do not, their typings will contain `any` where JSX references should be.

Setting these configurations ourselves, at least as defaults, allows us to make fewer assumptions about users' configs that are external to Microbundle (and hopefully will raise fewer issues).

src/index.js
```diff
@@ -549,9 +550,16 @@ function createConfig(options, entry, format, writeMeta) {
...
-									jsx: 'preserve',
-									jsxFactory: options.jsx,
-									jsxFragmentFactory: options.jsxFragment,
+									...(options.jsxImportSource
+										? {
+												jsx: 'react-jsx',
+												jsxImportSource: options.jsxImportSource,
+										  }
+										: {
+												jsx: 'preserve',
```
developit, line 559: should this be `preserve`, or `react`?

rschristian: `preserve` should be correct, though from my quick tests I don't think it makes a difference either way.

`preserve` is to be used when having another tool transpile JSX (e.g., babel), which is what we do. `react` on the other hand has TSC transpile it itself. [TS Docs](https://www.typescriptlang.org/docs/handbook/jsx.html#basic-usage)

I think, at best, `react` would do nothing, and at worst, it'd slow down type generation due to trying to handle JSX which we don't need. No one's raised any issues regarding `preserve` so I'd assume it's fine.

PR #992, opened 2022-08-23
Add example of exports.types field

### LINKED ISSUE #990: Guidance around exports.types config when authoring in TS
I'm authoring a module in TypeScript that I'm bundling and publishing with `microbundle`. I would like the published module to be importable from a wide range of projects, with the generated TypeScript definitions to be available as well.

The "[Using with TypeScript](https://github.com/developit/microbundle#using-with-typescript)" guide does not mention one piece of configuration that I've found necessary. In order to produce a module that can be imported from TypeScript projects with `moduleResolution: "NodeNext"` configured, with the `.d.ts` definitions available as well, I needed to add an [`exports.types` field](https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing) to [my module's `package.json` config](https://github.com/jeffposnick/remove-filename-hash/blob/d2f1086bafe9064bf38180468adb307842082222/package.json#L14-L18).

Without adding in that `types` field, I get the standard `ts(7016)` error when using my module from `moduleResolution: "NodeNext"` projects, informing me that there are no type definitions.

I'm happy to file a PR against the `microbundle` docs instructing developers to set `exports.types` for maximum compatibility with modern TypeScript.

jeffposnick:
Fixes #990 by adding some examples to the README, along with more context in the TypeScript section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@
   "type": "module",
   "source": "src/foo.js",             // your source code
   "exports": {
+    "types": "./dist/foo.d.ts",       // optional; only needed if authoring in TypeScript
     "require": "./dist/foo.cjs",      // used for require() in Node 12+
     "default": "./dist/foo.modern.js" // where to generate the modern bundle (see below)
   },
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,6 +42,7 @@
   "type": "module",
   "source": "src/foo.js",             // your source code
   "exports": {
+    "types": "./dist/foo.d.ts",       // optional; only needed if authoring in TypeScript
```
rschristian, line 45:
```suggestion
```

Sorry, I didn't realize we weren't previously documenting `"types"` in that example. That's my fault for suggesting this in the issue.

FWIW, with `--generateTypes`, Microbundle can create `.d.ts` files from the source, utilizing JSDoc.

Since `--generateTypes` defaults to `false`, and authoring in TS isn't a given, this was likely missing here to avoid references to files that may not exist.

rschristian (approved):

rschristian:
Thanks!

PR #993, opened 2022-08-25
Update README.md

arzafran:
Add missing punctuation and the libraries we've built that are using microbundle.


rschristian (approved):
Thanks!

PR #1043, opened 2023-05-06
fix: Incorrect UMD sourcemaps

### LINKED ISSUE #1041: Sourcemap mismatch for unpkg/UMD build
When building [wrld.js](https://github.com/wrld3d/wrld.js) for CDN consumption, either by passing `--format umd` to the `microbundle` command, or calling `microbundle` without arguments so it uses the `"unpkg"` entry in `package.json`, the source map produced contains incorrect offsets. The source maps produced for the `"main"`, `"exports"` and `"module"` builds work properly, though.

As far as I can tell, the issue can be worked around by adding forty four whitespace characters to the start of the emitted script, although I've not confirmed whether that works for everything or just the few things I tried. I don't see an obvious reason why there'd be this offset or why 44 is the number of extra characters that magically cancels it out.

rschristian:
Fixes #1041

`magic-string` is of course used by a ton of Rollup plugins, so this doesn't actually add anything to the dep tree.


package-lock.json
```diff
@@ -3260,6 +3274,14 @@
 				"magic-string": "^0.25.0"
 			}
 		},
+		"node_modules/@surma/rollup-plugin-off-main-thread/node_modules/magic-string": {
+			"version": "0.25.9",
```
ForsakenHarmony, line 3278: seems like this does actually add to the dep tree because you're using `0.30` while the existing one was `0.25`?

rschristian: Only affects us because we're using a rather ancient version in our lockfile; in an actual install, they'd resolve to the same version.

Do you prefer that I also bump the existing copy of `magic-string` in our lock file?

rschristian: Sorry if you saw the previous message, I realized I wasn't looking at what you had actually commented on.

Will take a peek, can probably drop the version used (& bump back to latest in that other PR)

rschristian:
Ugh, looks like it might rely on #1082 as the Compressed Size action is off by 1-2b

ForsakenHarmony (approved):

PR #1061, opened 2023-09-18
Add swup to example projects

daun:
Appending [our little project](https://github.com/swup/swup), in case you're still adding to the list :) Thanks for making this! We've been able to simplify our tooling immensely by adopting microbundle.


rschristian (approved):
Glad it helps!

PR #1071, opened 2024-01-05
Fixes anchor tags for NPM

schalkventer:
`<a src="#*">` are not supported on NPM, it might be better to just use the automatically generated anchor IDs since this means the table of contents will work on both Github and NPM

For example `https://www.npmjs.com/package/microbundle#formats` does not work, but `https://www.npmjs.com/package/microbundle#-output-formats-` works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,16 +11,16 @@
 
 <p align="center">
   <strong>Guide → </strong>
-  <a href="#setup">Setup</a> ✯
-  <a href="#formats">Formats</a> ✯
-  <a href="#modern">Modern Mode</a> ✯
-  <a href="#usage">Usage &amp; Configuration</a> ✯
-  <a href="#options">All Options</a>
+  <a href="#-installation--setup--">Setup</a> ✯
+  <a href="#-output-formats-">Formats</a> ✯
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,16 +11,16 @@
 
 <p align="center">
   <strong>Guide → </strong>
-  <a href="#setup">Setup</a> ✯
-  <a href="#formats">Formats</a> ✯
-  <a href="#modern">Modern Mode</a> ✯
-  <a href="#usage">Usage &amp; Configuration</a> ✯
-  <a href="#options">All Options</a>
+  <a href="#-installation--setup--">Setup</a> ✯
```
rschristian, line 14: This doesn't appear to work on GitHub though

rschristian: GitHub's markdown renderer is notoriously inconsistent cross-platform, but this doesn't work for me on Windows Chrome & FF, Linux FF, or iOS Safari.

Test on the readme of your fork, does this link actually work? https://github.com/schalkventer/microbundle/tree/patch-1#-installation--setup--

rschristian: I can't find a single environment in which those trailing `-`'s are present -- what's your system details? Where are you running this where that gets generated?

PR #1081, opened 2024-06-04
chore: Bump deps, mostly working

rschristian:
Supersedes #969


test/fixtures/class-decorators-ts/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
 	"compilerOptions": {
+		"moduleResolution": "node",
```
rschristian, line 3: `tslib` otherwise can't be found, oddly. This should be the default but maybe rtp2 is doing something odd.

package.json
```diff
@@ -68,56 +68,56 @@
...
-		"@babel/preset-flow": "^7.12.1",
-		"@babel/preset-react": "^7.12.10",
-		"@rollup/plugin-alias": "^3.1.1",
-		"@rollup/plugin-babel": "^5.2.2",
-		"@rollup/plugin-commonjs": "^17.0.0",
+		"@babel/plugin-syntax-jsx": "^7.24.6",
+		"@babel/plugin-transform-class-properties": "7.24.6",
+		"@babel/plugin-transform-flow-strip-types": "^7.24.6",
+		"@babel/plugin-transform-react-jsx": "^7.24.6",
+		"@babel/plugin-transform-regenerator": "7.17.9",
```
rschristian, line 77: TODO: somewhere after this results in regenerator ending up in the bundles, blowing them up rather significantly.

package.json
```diff
@@ -68,56 +68,56 @@
...
+		"tiny-glob": "^0.2.9",
 		"tslib": "^2.0.3",
 		"typescript": "^4.1.3"
 	},
 	"devDependencies": {
-		"@babel/cli": "^7.12.10",
-		"@babel/node": "^7.12.10",
-		"@babel/plugin-proposal-throw-expressions": "^7.12.1",
+		"@babel/cli": "^7.24.6",
+		"@babel/node": "7.12.10",
```
rschristian, line 116: TODO: Bumping this requires altering our build command to ensure the flags are passed to Microbundle, not `babel-node`

test/__snapshots__/index.test.js.snap
```diff
@@ -1856,30 +1855,30 @@ esnext-ts
...
-1011 B: esnext-ts.esm.mjs.gz
-896 B: esnext-ts.esm.mjs.br
-1072 B: esnext-ts.umd.js.gz
-959 B: esnext-ts.umd.js.br"
+1180 B: esnext-ts.js.gz
+1055 B: esnext-ts.js.br
+1181 B: esnext-ts.esm.mjs.gz
+1056 B: esnext-ts.esm.mjs.br
+1240 B: esnext-ts.umd.js.gz
+1112 B: esnext-ts.umd.js.br"
```
rschristian, line 1863: TODO: Sizable regression here, need to figure out what this is

package.json
rschristian: Oh nice, `7.18.0` ended up inlining regenerator as a normal helper, dropping the global: https://github.com/babel/babel/pull/14538

This is a sizable increase but IMO it's completely reasonable.

test/__snapshots__/index.test.js.snap
rschristian: Ah, that's [`AsyncFromSyncIterator`](https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/async-from-sync-iterator.js) -- going to call this fair enough.

.github/workflows/size.yml
```diff
@@ -16,10 +16,3 @@ jobs:
           pattern: 'dist/!(*.js.map)'
           build-script: build
           repo-token: '${{ secrets.GITHUB_TOKEN }}'
-
-      - name: compressed-size-action
-        uses: preactjs/compressed-size-action@v2
-        with:
-          pattern: 'test/fixtures/**/dist/!(*.map)'
-          build-script: 'test'
-          repo-token: '${{ secrets.GITHUB_TOKEN }}'
```
rschristian, line 25: Test suite already covers this really, we don't need a separate readout for file size bumps in a PR comment.

Also this is not aligning w/ the 'Test' CI action, probably Node version differences that isn't worth trying to hack around

hbroer:
This is IMO very important to merge. I need to manualy override "rollup-plugin-typescript2" to latest (0.36.0) everywhere I use microbundle, else it can't compile typescript because it can't find the types:

```
(rpt2 plugin) Error: ...ts(x,x): semantic error TS2307: Cannot find module '...' or its corresponding type declarations.
  There are types at '....d.ts',  but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
```

I had the solution found a year or so ago, but forgot about it and did get cought by it again and wasted a few hours to find out how to solve it. Workaround:

Add this to your `package.json`

```
  "overrides": {
    "rollup-plugin-typescript2": "^0.36.0"
  }
```

If you use npm workspaces make sure it is overwritten and not existing anymore in `node_modules/microbundle/node_modules`.

rschristian:
@developit If you have a few minutes, would appreciate a peek at this and potentially #1082 & #1083 -- we're starting to hit dep rot

src/lib/babel-custom.js
```diff
@@ -7,6 +7,16 @@ const ESMODULES_TARGET = {
 	esmodules: true,
 };
 
+// silence Babel 7.13+ loose-with-assumptions warnings
```
ForsakenHarmony, line 10: why do we want to silence this?

test/__snapshots__/index.test.js.snap
```diff
@@ -2320,26 +2319,26 @@ modern-generators
 
 
 Build \\"modern-generators\\" to dist:
-248 B: modern-generators.js.gz
-201 B: modern-generators.js.br
+2.57 kB: modern-generators.js.gz
```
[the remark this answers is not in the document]
rschristian: Hey, long time no see! Hope all is well & good w/ you.

Mentioned this above in a resolved comment:

> Oh nice, `7.18.0` ended up inlining regenerator as a normal helper, dropping the global: https://github.com/babel/babel/pull/14538
>
> This is a sizable increase but IMO it's completely reasonable. 

I still think this is reasonable, but admittedly, I was very much not a fan of Babel's previous behavior.

Do you disagree with this? I suppose we could work around it and revert to previous behavior if we really needed to.

src/lib/babel-custom.js
rschristian: Great question, will look into it & improve the comment (or remove this entirely) as I don't quite remember myself.

rschristian: Thanks for the mention! Indeed, seems like a silly hack I inserted when testing and forgot to clean up/fix.

test/__snapshots__/index.test.js.snap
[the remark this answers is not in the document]
ForsakenHarmony: All good, how about you?

So if I'm understanding it correctly, the previous size was somewhat misleading because the runtime was needed?
But now that it's inlined it will be duplicated for each library built this way

Maybe @developit could give his opinion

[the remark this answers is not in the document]
rschristian: Not bad, thanks!

You're right, yep. `regeneratorRuntime` was a unique case in that instead of being inlined, as every other Babel helper is, it was consumed as a global with the expectation that the user makes it available. This tripped up a great many users over the years ([Preact org issues](https://github.com/search?q=org%3Apreactjs+regeneratorRuntime&type=issues), for reference, and you can find many hundred similar issues elsewhere) as it was a bit odd and had questionable DX.

Indeed though, like all Babel helpers this can end up duplicated and it's particularly bad if it is -- even gzip isn't going to save you there.

That being said, I'm not sure there's a whole heck of a lot that we can do, or nothing comes to mind at least. If a user is bundling their deps with Microbundle, and one of those deps has this helper inlined, there's still duplication whether we stick with the old global or duplicate it. In a perfect world every module would be distributing modern bundles with minimal transpiling/helpers, but... :/ Heck, even our "modern" is still ES2017, could probably do with a bump to 2019-2020?

test/__snapshots__/index.test.js.snap
```diff
@@ -1194,30 +1193,30 @@ class-decorators-ts
...
+"\\"function\\"==typeof SuppressedError&&SuppressedError;var e=/*#__PURE__*/function(){function e(e){this.greeting=e}return e.prototype.greet=function(){return\\"Hello, \\"+this.greeting},e}(),t=new(e=function(e,t,r,o){var n,c=arguments.length,f=c<3?t:null===o?o=Object.getOwnPropertyDescriptor(t,r):o;if(\\"object\\"==typeof Reflect&&\\"function\\"==typeof Reflect.decorate)f=Reflect.decorate(e,t,r,o);else for(var l=e.length-1;l>=0;l--)(n=e[l])&&(f=(c<3?n(f):c>3?n(t,r,f):n(t,r))||f);return c>3&&f&&Object.defineProperty(t,r,f),f}([function(e){Object.seal(e),Object.seal(e.prototype)}],e))(\\"Hello World\\");export default t;
```
developit, line 1207: this is weird. I wonder why it's being injected? it's dead code.

test/__snapshots__/index.test.js.snap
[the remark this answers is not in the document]
developit: I am wondering why the modern output has regenerator in it at all, since ES2017 supports generators.

Maybe we need to [configure Babel `"assumptions"`](https://babeljs.io/docs/assumptions#migrating-from-babelpreset-envs-loose-and-spec-modes)?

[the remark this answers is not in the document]
rschristian: `modern-generators` is the package name (tripped me up too), `modern-generators.modern` is indeed unchanged in this PR

test/__snapshots__/index.test.js.snap
rschristian: Comes from [`tslib`](https://github.com/microsoft/tslib/blob/d72d6f70b36286bc3f94a3dda1e64dcb568b1370/tslib.es6.mjs#L328) which seems to treat it like a global. Not quite sure why terser keeps it around.

developit: It looks like we can strip this out via `compress.global_defs`:

```
compress: {
  global_defs: {
    SuppressedError: 'undefined'
  },
}
```

Screenshot 2025-01-07 at 8 59 13 PM

PR #1082, opened 2024-06-04
ci: Bump CI versions

rschristian:


.github/workflows/size.yml
```diff
@@ -21,5 +21,6 @@ jobs:
         uses: preactjs/compressed-size-action@v2
         with:
           pattern: 'test/fixtures/**/dist/!(*.map)'
-          build-script: 'test'
+          # We don't care about test suite failures, that's for the CI action
+          build-script: 'test -- -u'
```
rschristian, line 25: Perhaps questionable, but https://github.com/developit/microbundle/pull/1081/commits/fe9728619c8344252a77e161f20844695ea833bf shows a failing compressed-size action when the real 'Test' action passes just fine.

We shouldn't care about test failures here though.

rschristian:
Bah, looks like `gzip-size` might not be the most stable output.

.github/workflows/nodejs.yml
```diff
@@ -3,36 +3,24 @@ name: Node CI
 on:
   pull_request: {}
   push:
-    branches-ignore:
-      - trying.tmp
-      - staging.tmp
+    branches:
+      - master
```
rschristian, line 7: Limiting to master else a PR gets a CI run from both `push` and & `pull_request` triggers

.github/workflows/size.yml
```diff
@@ -8,18 +8,11 @@ jobs:
...
           pattern: 'dist/!(*.js.map)'
           build-script: build
           repo-token: '${{ secrets.GITHUB_TOKEN }}'
-
-      - name: compressed-size-action
-        uses: preactjs/compressed-size-action@v2
-        with:
-          pattern: 'test/fixtures/**/dist/!(*.map)'
-          build-script: 'test'
-          repo-token: '${{ secrets.GITHUB_TOKEN }}'
```
rschristian, line 25: I don't know that this is particularly useful, any size difference will be accounted for in the tests already.

test/index.test.js
```diff
@@ -43,7 +42,7 @@ describe('fixtures', () => {
 
 			await sleep(1);
 
-			const output = await buildDirectory(fixtureDir);
```
rschristian, line 46: Unfortunately we need to drop this aspect of the suite. gzip & brotli sizes aren't guaranteed to be stable, not through Node versions or on different systems. As such, I was seeing tests pass locally just fine that would then be 1-2b off on the CI.

Besides, all that this aspect was testing was that the gzip & brotli size estimates were stable, which isn't particularly important. If the expected output is stable, it doesn't really matter to us what the gzip/brotli size ends up being.

rschristian:
Will then need the required CIs to be updated

test/index.test.js
ForsakenHarmony: hmmm do we get any reporting on the size then?
I guess compressed size action?

rschristian: Compressed Size, yeah.

It's not ideal but I don't have any other ideas for how to deal with this.

ForsakenHarmony (approved):

test/index.test.js
rschristian: Well, as mentioned, compressed-size will report this for PRs now, so there will be some visibility yet.

Certainly not ideal but I don't know what else to do there, I couldn't manage to get local to align with CI due to different system paths and the like in the bundles/sourcemaps.

rschristian: Well, strangely the sizes align on both the CI (Node 12.x & 14.x) as well as my local (Node 22.19) for the moment, so this was all reverted.

Maybe something got ironed out? Dunno.

PR #1086, opened 2024-08-05
Update the path to the modern bundle in the example in README.md

mlshv:
In the example path to modern bundle ends with .js, but microbundle generates .mjs file


rschristian:
That example has `"type": "module"` which means it'll be generated as `.js` actually. It's correct as-is.

PR #1091, opened 2025-02-16
fix: Resolve entries against cwd

rschristian:
https://github.com/preactjs/preact/pull/4677

[`tiny-glob` resolves against `opts.cwd` or `.`](https://www.runpkg.com/?tiny-glob@0.2.9/sync.js#56), and because we don't pass `cwd` into `tiny-glob`, this can result in incorrect file paths when used with Microbundle's `--cwd` flag.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -218,7 +218,7 @@ async function getInput({ entries, cwd, source, module }) {
 						(await jsOrTs(cwd, 'index')) ||
 						module,
 		)
-		.map(file => glob(file))
+		.map(file => glob(file, { cwd }))
 		.forEach(file => input.push(...file));
 
 	return input;


tools/build-fixture.js
```diff
@@ -56,6 +56,10 @@ export const buildDirectory = async fixtureDir => {
 	await rimraf(resolve(`${fixturePath}/.rts2_cache_es`));
 	await rimraf(resolve(`${fixturePath}/.rts2_cache_umd`));
 
+	// Now that dirs have been cleared, reset the fixturePath back to the parent dir
+	if (fixtureDir.endsWith('-with-cwd')) {
+		fixturePath = resolve(FIXTURES_DIR, fixtureDir);
+	}
```
rschristian, line 62: This was super broken & went unnoticed for a long time.

Basically we make the assumption that a test fixture `foo-with-cwd` is testing `--cwd ./foo`, and to make sure we were clearing out past runs, set the fixture path to `fixtures/foo-with-cwd/foo` so that `rimraf` would clear out all test cases the same way. However, we never set this path back! 

When it came time to get the build script & execute it, the CWD was `/foo-with-cwd/foo`... none of these nested directories actually had a build script defined, as it was meant to come from the parent, and so all of the `-with-cwd` tests were just running plain `microbundle`. The build script we provided for our tests was entirely ignored as the test was always ran from the incorrect working dir.

Easy fix though, we just have to reset it.

